PR #140, opened 2015-01-21
TESB-14779: It's not possible to create a new job from ctalendjob into s...

csuntalendbj:
pls merge to tcommon-studio-se


ggu-talend:
I think, should find another way to fix this bug, because it work for case insensitive in win for  TDI-29841.

ggu-talend:
Yes, after check the fix in TDI-30470, it has been removed.

PR #144, opened 2015-02-03
Spark ga refactoring

rdubois:


nrousseau:
just note that it's better to remove the jars from the core.runtime plugin in the final commit

PR #177, opened 2015-04-17
TDI-32083: place the connection file into configuration again

sgandon:
Please can you test this before merging into master, details of the test to be done are in https://jira.talendforge.org/browse/TDI-32083
Thanks.


sgandon:
this was done manually by our chinese colleagues.

PR #189, opened 2015-05-13
TUP-2674: Upgrade CXF jar to v3.1 in tWebService

lwang-talend:


amarkevich:
Can be existing dependencies used for this component?
<IMPORT MODULE_GROUP="cxf-rt-frontend-jaxws" REQUIRED="true" />

amarkevich:
CXF part of org.talend.libraries.esb can be refactored to be org.talend.libraries.apache.cxf

nrousseau:
org.talend.libraries.apache.cxf is included in every builds from TOS, so yes, you can reuse those jars.

nrousseau:
FYI, merged in another pull request.

PR #190, opened 2015-05-13
TDQ-10298 survivship package name could contain "." and "_".

qiongli:
Please review this change,if it is ok, please merge it to Talend master.
Thanks


ggu-talend:
We can use type to check for docs only.

ggu-talend:
Have send one email with git patch for this.

PR #293, opened 2015-11-23
Fix the problem: enclosed schema names will not be found.

jlolling:
The wizard fails with the error message "Schema not present in database".
The problem is caused by the quotas which comes from the GUI, but the getSchema() method in the DatabaseMetaData API never returns names with quotas.


jlolling:
Because I have to recreate my forked repo I close this pull request.

PR #298, opened 2015-11-26
 [TBD-2773] Empty "COMPONENT_LIST" element are now displaying an error

pbailly:
Without this fix, the component need a field "USE_EXISTING_COMPONENT", otherwise it will never display any error.


pbailly:
This pull request is invalid, please do not merge it.

PR #313, opened 2016-02-26
TUP-3842: fix compile errors.

ycbai:


sgandon:
I never understood why did you have to implement the service interface in the Studio, can you explain the reason ?

sgandon:
I have to accept this because the studio will fail tonight but I really would like to know why you have to implement this method. It is not supposed to be implemented and you put yourself in a position to have complication errors even when non breaking changes are done.

PR #326, opened 2016-03-11
ITDI-35132:tAccessInput throws OOM when retrieve tables with big size

hwang-talend:
over 100MB


ggu-talend:
It's ok to enable set additional parameter for Access.

PR #341, opened 2016-03-23
[TCOMP-87] fixed the dynamic column issue

bchen-talend:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -165,11 +165,16 @@
...
         // Types with unknown elements, store as binary
-        if (JavaTypesManager.OBJECT.getId().equals(tt) || JavaTypesManager.DYNAMIC.getId().equals(tt)) {
+        if (JavaTypesManager.OBJECT.getId().equals(tt)) {
             return defaultValue == null //
             ? ftb.bytesType().noDefault()
                     : ftb.bytesType().bytesDefault(defaultValue);
         }
+
+        if ("id_Dynamic".equals(tt)) { //$NON-NLS-1$
+            return ftb.nullType().noDefault();
```
RyanSkraba: I think this is better than bytes :)  Is it a problem if a Dynamic column is nullable?

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -178,6 +202,75 @@
         throw new UnsupportedOperationException("Unrecognized type " + tt); //$NON-NLS-1$
     }
 
+    private static Schema copyDynamicColumnProperties(Schema schema,
```
RyanSkraba: I had forgotten all about these properties that also need to be maintained.  Good catch.

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -178,6 +202,75 @@
...
+        }
+        for (TaggedValue tv : in.getTaggedValue()) {
+            String additionalTag = tv.getTag();
+            if (tv.getValue() != null) {
+                schema = AvroUtils.setProperty(schema, Talend6SchemaConstants.TALEND6_DYNAMIC_ADDITIONAL_PROPERTIES
+                        + additionalTag, tv.getValue());
+            }
+        }
+
+        // Column-specific properties.
```
RyanSkraba: Just a quick note: none of these properties need to be retained for the dynamic column, do they?

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -299,10 +392,86 @@ public static MetadataTable convertFromAvro(org.apache.avro.Schema in) {
...
+        }
+        for (String key : schema.getJsonProps().keySet()) {
+            if (key.startsWith(Talend6SchemaConstants.TALEND6_ADDITIONAL_PROPERTIES)) {
+                String originalKey = key.substring(Talend6SchemaConstants.TALEND6_ADDITIONAL_PROPERTIES.length());
+                TaggedValue tv = TaggedValueHelper.createTaggedValue(originalKey, schema.getProp(key));
+                col.getTaggedValue().add(tv);
+            }
+        }
+
+        // Column-specific properties.
```
RyanSkraba: As above, aren't all of the properties below constant for dynamic columns?  I really don't know!

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -178,6 +202,75 @@
...
+        if (in.isKey()) {
+            schema = AvroUtils.setProperty(schema, Talend6SchemaConstants.TALEND6_COLUMN_IS_KEY, "true"); //$NON-NLS-1$
+        }
+        if (in.getType() != null) {
+            schema = AvroUtils.setProperty(schema, Talend6SchemaConstants.TALEND6_COLUMN_SOURCE_TYPE, in.getSourceType());
+        }
+        if (in.getTalendType() != null) {
+            schema = AvroUtils.setProperty(schema, Talend6SchemaConstants.TALEND6_COLUMN_TALEND_TYPE, in.getTalendType());
+        }
```
bchen-talend: here for id_Dynamic

PR #345, opened 2016-03-24
TUP-4273 Add jackson and avro bundles where necessary.

RyanSkraba:


main/plugins/org.talend.core.runtime/META-INF/MANIFEST.MF
```diff
@@ -111,7 +111,8 @@ Require-Bundle: org.eclipse.jdt.core,
  org.talend.commons.ui,
  org.ops4j.pax.url.mvn,
  org.talend.components.api,
- org.apache.servicemix.bundles.avro
+ org.apache.servicemix.bundles.avro,
+ jackson-core-asl;bundle-version="1.9.13"
```
sgandon: we try to avoid specifying the version in the Manifest to allow for easy update.
could you remove those please, along with the avro one ?

RyanSkraba:
Fixed version info.

PR #349, opened 2016-03-30
complete built-in Avro type

bchen-talend:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -134,72 +132,62 @@
...
+            defaultValue = defaultValue == null ? null : Double.parseDouble(defaultValue.toString());
+        } else if (JavaTypesManager.FLOAT.getId().equals(tt)) {
+            type = AvroTypes._float();
+            defaultValue = defaultValue == null ? null : Float.parseFloat(defaultValue.toString());
+        } else if (JavaTypesManager.BIGDECIMAL.getId().equals(tt)) {
             // decimal(precision, scale) == column length and precision?
-            Decimal d = LogicalTypes.decimal((int) in.getLength(), (int) in.getPrecision());
-            Schema bigdecimal = d.addToSchema(Schema.create(Schema.Type.BYTES));
-            return defaultValue == null ? fb.type(bigdecimal).noDefault() : fb.type(bigdecimal).withDefault(defaultValue);
+            type = AvroTypes._decimal();
```
RyanSkraba: OK .. I see a problem here:

There's a difference between BigDecimal when the scale and precision is always the same for every value (like Salesforce) where we should be using the LogicalTypes.decimal, and when the scale and precision isn't guaranteed to be the same for every value (like Cassandra) where we don't have any choice but to use String.

I'm not sure exactly what we should be doing here in tcommon... 

Any suggestions?  Should we just enforce the use of String for all cases?

bchen-talend: Checked the document of [BigDecimal](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#toString--), it said use toString() and BigDecimal(String) will get the original data, so it won't lose anything, so seems we don't need to keep the information of precision and scale, and we also have TALEND_COLUMN_PRECISION and TALEND_COLUMN_SCALE, so seems ok then.
What do you think?

PR #353, opened 2016-04-01
TUP-4234: Do not restart studio if only jars were downloaded from additional packages.

zwxue:


main/plugins/org.talend.updates.runtime/src/main/java/org/talend/updates/runtime/model/TalendWebServiceUpdateExtraFeature.java
```diff
@@ -254,4 +254,9 @@ public ExtraFeature createFeatureIfUpdates(IProgressMonitor progress) throws Exc
         return null;// we do not handle udpates yet.
     }
 
+    @Override
+    public boolean needRestart() {
+        return false;
```
nrousseau: this is false for DQ with their implementation of TalendWebServiceUpdateExtraFeature, which extends this class. It should be true by default here also. (what they download is plugins, so it must require a restart)

nrousseau:
accept still, the commit on DQ side will come in a few

PR #358, opened 2016-04-06
EXASol JDBC driver update

jlolling:
We should use the driver version 5.0.13 now.


pteyssier:
Note: We will not integrate the jar through git.

PR #381, opened 2016-04-19
TUP-4369: add junits

ycbai:


nrousseau:
why not full revert of the commits already done on test utils? like required bundles of test utils?

PR #391, opened 2016-04-22
TMDM-9493 : Move dependency jackson libraries for tMDMRestInput to or…

lpw82gf:
…g.talend.libraries.jackson


amarkevich:
can be com.fasterxml.jackson.jaxrs.json package exported to be used in other plugins?

glzhou:
It certainly can, but it would better to handle the package exporting with a separate PR, linked to your task. We don't know if exported package may or may not causing classloading issue if there are other versions of same package being used in the studio.

PR #414, opened 2016-05-03
TESB-17272: Fixed cannot open route test case editor

yanbasic:


main/plugins/org.talend.repository.view/src/org/talend/repository/viewer/action/RepoDoubleClickAction.java
```diff
@@ -405,7 +405,7 @@ private ITreeContextualAction getAction(RepositoryNode obj) {
                     return current;
                 }
 
-            } else if (nodeType != null && nodeType.equals(ERepositoryObjectType.PROCESS_ROUTE)
+            } else if (nodeType != null && ERepositoryObjectType.PROCESS_ROUTE.equals(obj.getContentType())
                     && isDITestCaseEditOrReadAction(current)) {
```
nrousseau: Actually PROCESS_ROUTE can be null for every product without routes (like TOS_DI etc...).
So lots of chances to get a NPE here.

PR #430, opened 2016-05-12
TDI-35569:zoom out within job designer does not work as expected in

nrousseau:
Studio


main/plugins/org.talend.commons.ui/src/main/java/org/talend/commons/ui/gmf/draw2d/AnimatableZoomManager.java
```diff
@@ -35,6 +35,8 @@
 
     private int zoomAnimationStyle = ANIMATE_NEVER;
     private List animationListeners = new ArrayList();
+    
+    public static double currentAoom = 1.0;
```
nrousseau: please use correct naming at least

PR #439, opened 2016-05-17
TDI-36029: fix the compile error of ComponentServiceWithValueEvaluator

ycbai:


nrousseau:
studio actually still take the old version of api in the build, need to modify on the build configuration first to take the correct version before merge this

ycbai:
It is no need since I will remove ComponentServiceWithValueEvaluator after.

PR #447, opened 2016-05-25
TDI-36027：trunjob component throwing no class found error

hwang-talend:


ggu-talend:
Ok to remove the dependences for children jobs

PR #477, opened 2016-06-08
TBD-3563:improve mapr tickets support

hcyi:


main/plugins/org.talend.metadata.managment/src/main/java/org/talend/metadata/managment/connection/manager/HiveConnectionManager.java
```diff
@@ -218,6 +167,45 @@ private Connection createHiveStandloneConnection(IMetadataConnection metadataCon
         return hiveStandaloneConn;
     }
 
+    private void setMaprTicketConfig(IMetadataConnection metadataConn, ClassLoader classLoader, boolean useMaprTicket,
+            boolean useKerberos) throws SQLException {
+        if (useMaprTicket) {
```
ycbai: This condition should be out of the method since there is not any operation when it is false.

ycbai:
It's OK.

PR #485, opened 2016-06-12
Tdi 36031 slowing performance after 100 jobs

wdliu:
TDI-36031：Slowing Performance in Studio After 100 jobs


ggu-talend:
I prefer to understand why refresh several times. What's happen? which resources was changed? etc.

nrousseau:
Issue was, content provider of:
- job design
- overrided
- test case

are all triggered each time there is a change done on any job.

One possible issue i think:
- import items, after this many refresh should be triggered from the different types
=> only some refresh will by done by luck, since we can totally have one refresh of job / metadata done at the same time in theory.

Or what we need here is a pool of refresh... since in all case even the repository doesn't fully support multithread.

PR #493, opened 2016-06-15
TUP-4175：Export with dependencies fails to select (and export)

hwang-talend:
dependencies

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
@@ -534,12 +534,17 @@ public class RelationshipItemBuilder {
                     return o1.getType().compareTo(o2.getType());
                 }
             });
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
```diff
@@ -520,15 +534,25 @@ public int compare(Relation o1, Relation o2) {
                     return o1.getType().compareTo(o2.getType());
                 }
             });
-            for (Relation relatedItem : relationItemsList) {
+            out:for (Relation relatedItem : relationItemsList) {
+                List<ItemRelation> relationList = new ArrayList<ItemRelation>(itemRelations.getRelatedItems());
+                for(ItemRelation item : relationList){
+                    if(item.getId().equals(relatedItem.getId())){
+                        continue out;
```
nrousseau: Please, NEVER use this.
Use standard loop  / break.
This is impossible to understand when you read the code

nrousseau:
This part is critical enough in the product to need some junits here.. please add some

PR #507, opened 2016-06-24
Bugfix/6.2.1/tdi 35810 when log on main project

wdliu:
DITDI-35810: When log on Main project can not open referrence project's system routines.


main/plugins/org.talend.core.runtime/src/main/java/org/talend/repository/model/RepositoryNodeUtilities.java
```diff
@@ -79,6 +79,10 @@ public static IPath getPath(final IRepositoryNode node) {
         if (node.isBin()) {
             return new Path(""); //$NON-NLS-1$
         }
+        String systemNodelabel = node.getLabel();
+        if (node.getType() == ENodeType.STABLE_SYSTEM_FOLDER && systemNodelabel.equals("system")) {
```
ggu-talend: better use same codes like " if (RepositoryConstants.SYSTEM_DIRECTORY.equalsIgnoreCase(node.getLabel()))"

PR #540, opened 2016-07-13
TESB-17301: Open Spring view at the bottom area

yanbasic:
Hello, I added the Spring View Id as a placeholder in the bottom area, for the '*' placeholder is not work. The Spring View is for ESB route.

Thanks,


ggu-talend:
Seems from the eclipse JDT or such, use this "addPlaceholder" API too.
so can do it like this. and merge still.

PR #557, opened 2016-07-21
Workitem/6.3/tup 4478 move talend6 from daikon

gonchar-ivan:
Related to https://jira.talendforge.org/browse/TUP-4478

Please, check it carefully. It was my first time adding eclipse plugin.
I'm not very good with building process.


sgandon:
I wonder why you removed the .project and .classpath which are the eclipse metadata that we usually store in plugin projects ?

sgandon:
also the name of the bundle is not very meaningful...
what do you think of 
    org.talend.schema.enforcer
or
    org.talend.generation.enforcer

gonchar-ivan:
I like org.talend.schema.enforcer name.
I will rename.

I thought that it is wrong to store .project and .classpath in repository, because these files could contain developer specific settings (e.g. classpathenty, which points to some directory on developer machine). Also I don't know whether these files are required.

I will restore them, it they are required

gonchar-ivan:
Made changes according Sebastien's comments

gonchar-ivan:
@nrousseau, could you also review it?

nrousseau:
@sgandon  I thought, for any new plugin / package name, there should be some approval first.
So not sure the plugin name is really good here.

Anyway this is only one part of course.
In the next steps there will be at least:
- remove the DiSchemaConstants from org.talend.daikon.di
- remove the DiSchemaConstants from org.talend.core.runtime (that was similar to this task here, but done maybe too early...)
- add the plugin here to an existing feature for the studio build
- add the plugin as exported by default for any job
- setup the pom.xml to build with java 7 instead of java 8 (to be compatible with jobs)
- update the Component class to add this jar by default in the classpath for every job built.

I guess we can do all of this in this same branch... no need to merge directly now.
However not sure i can get people to help you on this right now... I think you will have to wait a little.

gonchar-ivan:
This is not relevant anymore.
This library will be used instead.
https://github.com/Talend/tdi-studio-se/tree/master/main/plugins/org.talend.designer.components.libs

PR #571, opened 2016-08-04
TDI-36954 bug solved

jlolling:
This is the bug fix to the problem TDI-36954.
Thge changes are in the plugin: org.talend.metadata.management 
in the class: org.talend.metadata.managment.model.DBConnectionFillerImpl.java
in the methods: fillColumns
I have commented the changed lines.


jlolling:
Obviously merged. It would be great to know from the Talend developers if they actually do not want pull requests - mostly they are ignored.

PR #579, opened 2016-08-11
TBD-3744 fix: Mapr 410&401: Missing Authentication option on hive wizard

guol-talend:
https://jira.talendforge.org/browse/TBD-3744


ycbai:
The logic is not right, please redo it.

PR #585, opened 2016-08-12
TMDM-9884 : [DI] TDS - Implement widget of Campaign Chooser for tData…

lpw82gf:
…StewardshipTaskOutput


nrousseau:
Why do we need an ElementParameterType?
It would be better to have only the common types here.
We should be able to use extension points only normally (or it makes no points to have the extension points if we need to edit an enum)

nrousseau:
forget the node above, will merge still... i created this first to improve on studio side: https://jira.talendforge.org/browse/TUP-5170

PR #586, opened 2016-08-15
Bugfix/6.2/tbd 3744 mapr 410&401 missing authentication option on hive wizard

guol-talend:
do some update


guol-talend:
wrong commit / pull request

PR #588, opened 2016-08-15
Bugfix/6.3.0/tdi 37032 can not build job

wchen-talend:


ggu-talend:
Maybe need do same checking for job/project dependencies also, ProcessorDependenciesManager.updateDependencies line 53

PR #609, opened 2016-08-24
TDI-37100:Create a IMDMService class(TDI-37001)

hcyi:


hcyi:
i will merged this to master directly , MDM will implement it today .

PR #620, opened 2016-08-31
TDI-37189:TestCase: No History show after running testcase on new

hwang-talend:
created testcase


main/plugins/org.talend.core/src/main/java/org/talend/designer/runprocess/ProcessorUtilities.java
```diff
@@ -1751,5 +1737,17 @@ private static IProcessor findProcessorFromJobList(String processId, String cont
...
+    private static void setAutoBuilding(boolean isAuto){
+        IWorkspace workspace = ResourcesPlugin.getWorkspace();
+        IWorkspaceDescription desc = workspace.getDescription();
+        desc.setAutoBuilding(isAuto);
+        try {
+            workspace.setDescription(desc);
+        } catch (CoreException e) {
+            CommonExceptionHandler.warn(e.getMessage());
+        }
+        ResourcesPlugin.getWorkspace().getDescription().isAutoBuilding();
```
nrousseau: this line can be removed.. it was only for testing

PR #631, opened 2016-09-05
TMDM-9952: add requirement of bundle jackson-databind-2.5.3.jar,jacks…

sbliu:
…on-module-jaxb-annotations-2.5.3.jar, make reexport, do not explicitly specify export related package again, to avoid jackson related export package conflict.


glzhou:
Cannot reproduce the original issue in latest 6.3.0 SNAPSHOT, pending investigation.
There is also prerequisite step to update P2 repository with 2.5.3 jackson jars.  Please do not merge.

glzhou:
PR is no longer needed, the function is working without issue.

PR #655, opened 2016-09-20
TMDM-9764 : Give the ability to export as a context a Talend MDM conn…

lpw82gf:
…ection


glzhou:
working on the fix...

PR #664, opened 2016-09-22
Feature/6.3/tdi 36957 file delimited

jzhao-talend:


jzhao-talend:
Need to merge after https://github.com/Talend/components/pull/249 merged

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -184,7 +184,7 @@ else if (JavaTypesManager.STRING.getId().equals(tt) || JavaTypesManager.FILE.get
 
         // Types with unknown elements, store as binary
         if (JavaTypesManager.OBJECT.getId().equals(tt)) {
-            // FIXME it's not right, as it don't store all the information about the object
+            type = AvroUtils._string();
```
wwang-talend: i think this is a temp fix, now we can use it.

wwang-talend:
can be merged

PR #680, opened 2016-10-12
TDI-37525: Add support document type in new framework

jzhao-talend:
https://jira.talendforge.org/browse/TDI-37525


wwang-talend:
it's ok.

PR #685, opened 2016-10-14
TMDM-9925: Add new fragment,to split mdm webservice jar for respectiv…

sbliu:
…ely CE and EE mdm studio.


glzhou:
PR No longer needed as TOS patch is provided from JIRA attachement

PR #697, opened 2016-10-20
TMDM-9925: Add a fragment to patch a CE webservice jar to host plugin.

sbliu:


main/plugins/org.talend.libraries.mdm/build.properties
```diff
@@ -4,4 +4,5 @@ bin.includes = META-INF/,\
                lib/commons-discovery-0.2.jar,\
                lib/mdm_webservice_tomcat.jar,\
                lib/xmlschema-core-2.2.1.jar,\
-               plugin.xml
+               plugin.xml,\
+               lib/mdm_webservice_tomcat_ex.jar
```
glzhou: why the naming is not consistent with mdm_webservice_tomcat_ext.jar?
better be explict about the what the jar is, it's Community Edition jar, so mdm_webservice_tomcat_ce.jar is a better name.

hbhong:
Validated.

main/plugins/org.talend.libraries.mdm/META-INF/MANIFEST.MF
```diff
@@ -20,6 +20,7 @@ Bundle-ClassPath: .,
  lib/stax2-api-3.1.1.jar,
  lib/woodstox-core-asl-4.4.1.jar,
  lib/xmlschema-core-2.2.1.jar,
+ lib/mdm_webservice_tomcat_ce.jar,
  lib/mdm_webservice_tomcat.jar
```
glzhou: when fragment org.talend.libraries.mdm.webservice.ce is included (in TOS studios), the mdm_webservice_tomcat_ce.jar will take precedence. while in EE studio, fragment org.talend.libraries.mdm.webservice.ce is not available, mdm_webservice_tomcat.jar (which is the EE version) will be loaded. 

So it's OK.

PR #702, opened 2016-10-21
TUP-16076 fix:HDFS file browser is not open when using hdfs component on

guol-talend:
CDH 58 Ker with HA


ycbai:
No change is needed. So close it.

PR #726, opened 2016-11-10
TUP-16072: Compiler error happen when Main job call a Big Data Batch.

zwxue:


nrousseau:
close since it will be fixed in the refactor done for CI

PR #729, opened 2016-11-11
TUP-16088:Rework SSL preferences (DataPrep+MDM)

wchen-talend:


glzhou:
Please be aware that SSLContextProvider is reused by TDS component designer, refactoring will impact TDS components

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/SSLContextProvider.java
```diff
@@ -39,21 +43,9 @@
...
-        @Override
-        public void checkServerTrusted(X509Certificate[] xcs, String string) throws CertificateException {
-        }
-
-        @Override
-        public X509Certificate[] getAcceptedIssuers() {
-            return null;
-        }
-    };
+    private static final IPreferenceStore store = CoreRuntimePlugin.getInstance().getCoreService().getPreferenceStore();
```
glzhou: I think introduce this dependency will not allow it to be reused outside of Studio plugin environment. i.e. TDS component bundle cannot get preference

wchen-talend: Move the SSLContextProvider.java to daikon will be don in TDKN-90.
This PR will be closed and new PR will be created to make TDKN-90 more easy to do.

PR #732, opened 2016-11-14
TUP-16088:Rework SSL preferences (DataPrep+MDM)

wchen-talend:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/SSLContextProvider.java
```diff
@@ -1,126 +1,117 @@
...
+
+            KeyStore tks = KeyStore.getInstance(keytype);
+            tks.load(stream, storePass.toCharArray());
+
+            KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509"); //$NON-NLS-1$
+            kmf.init(tks, storePass.toCharArray());
+
+            return kmf.getKeyManagers();
+        } catch (Exception e) {
+            e.printStackTrace();
```
pbabin: Forbid usage of printStackTrace, use of logging

PR #740, opened 2016-11-17
TDI-37741: Automatic Type Conversion from Object to byte/short/int/long/..... issue

jzhao-talend:
https://jira.talendforge.org/browse/TDI-37741


jzhao-talend:
This pull request should merged before https://github.com/Talend/tdi-studio-se/pull/805

PR #754, opened 2016-11-23
TDI-37709 Date/Time issues

francisu:
See corresponding change in Daikon. No tests provided because there is not a unit test framework for this as far as I know.


gonchar-ivan:
Seems, there are tests for tcommon plugins:

https://github.com/Talend/tcommon-studio-se/blob/master/test/plugins/org.talend.core.runtime.test/src/org/talend/core/model/metadata/MetadataToolAvroHelperTest.java

However, Nicolas said that I need to setup a lot of tools, if I want to implement tests for Studio plugins.
So, it is complicated for us and could be omitted.

gonchar-ivan:
LGTM!

However, what about convertToAvro() method?

francisu:
@gonchar-ivan I have reworked this to provide the conversion the other way (in fact I would not have approved this request as it was). I also took the approach of adding information to the DI types (the tagged values) so that we can get back to the right Avro type from DI. I think we need to migrate all of the dates to the Avro logical types at some point (after 6.3.1 of course) - see https://jira.talendforge.org/browse/TCOMP-370. I welcome your comments. This change also depends on the corresponding branch in Daikon.

gonchar-ivan:
Yes, that's what I meant )
Thanks for this change!

PR #760, opened 2016-11-25
TUP-5941 fix:Studio throw an error log when click finish button for a

guol-talend:
jdbc-access connection.


nrousseau:
we really can't have this kind of code here... better find another solution

PR #765, opened 2016-11-27
TUP-16465 NPE in handling Avro logical types

francisu:


nrousseau:
checked with junits, no problem.
I didn't know those functions for avro, good to learn new things !

francisu:
How can I run the junits? Ivan said it was hard.

nrousseau:
for most indeed yes, because it needs other plugin dependency, so a full target platform, and only RCP devs really know how to setup this after.
Luckily here, for this test, MetadataToolAvroHelperTest have no dependency, means it only needs to have the plugin org.talend.core.runtime.test, then right click on the class : Run As Junit.

PR #772, opened 2016-12-01
TDI-36295 : data type mapping: wrong DB->Java mapping

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-36295


wwang-talend:
LGTM

PR #789, opened 2016-12-12
bugfix TUP-15924 fix:studio GUI problem when open schema dialog

guol-talend:


main/plugins/org.talend.common.ui.runtime/src/main/java/org/talend/commons/ui/runtime/swt/tableviewer/TableViewerCreatorNotModifiable.java
```diff
@@ -689,12 +689,23 @@ protected int checkTableStyles() {
...
         }
         return style;
     }
 
+    private boolean isSupportOS() {
+        String osName = System.getProperty("os.name").toLowerCase();
+        String osVersion = System.getProperty("os.version");
+        Boolean isLinux = osName.startsWith("linux");
+
+        if (WindowSystem.isGTK() && isLinux && osVersion != null && "3.2.0-23-generic".equals(osVersion)) {
```
nrousseau: you of course can't base on one OS.
Any linux with the same configuration of libs would have the same.
Better have a system property like "disableLazyLoading" and if set to true we won't set virtual here. (user would set it manually)

PR #794, opened 2016-12-12
feat(TCOMP-230): add FlowVariablesReader available for codegen

gonchar-ivan:
Related to https://jira.talendforge.org/browse/TCOMP-230

This class is a Decorator(Wrapper) for TCOMP Reader. It adds additional code for getCurrent()
This additional code handles flow variables (as it was discussed in scope of TCOMP-230)
I need this class available in classpath of generated code, when TCOMP components are used. 
So, I decided to add it in routines.
Maybe there is a better place for it?

Another question is:
Is it ok that this class depends on components-api and components-common

### DIFF
--- a/main/plugins/org.talend.librariesmanager/resources/java/routines/system/FlowVariablesReader.java
+++ b/main/plugins/org.talend.librariesmanager/resources/java/routines/system/FlowVariablesReader.java
@@ -1,0 +1,180 @@
+package routines.system;
+
+import java.io.IOException;
+import java.util.Map;
+import java.util.NoSuchElementException;
+
+import org.apache.avro.Schema;
+import org.apache.avro.Schema.Field;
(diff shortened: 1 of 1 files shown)


gonchar-ivan:
@nrousseau Could you please check it?

nrousseau:
a jar might be better than a routine for now.
Problem with a system routine:
- to compile we must have all jars in the classpath
- this routines needs normally joda-time / avro / components.api, and need normally to update plugin.xml to set the relationship to the needed jars to compile.
- as soon as you set the plugin.xml dependency, every jobs, even a job with only a tMsgBox will need to include every system routines dependency. (we can't for now include routine only if some components are used)

If you need some other jar dependency to be added for every components of the new framework, you can add it simply in: org.talend.designer.core.generic.model.Component.getModulesNeeded(), same as we do right now for slf4j.

PR #806, opened 2016-12-19
Bugfix/6.4.1/tup 14737 too slowly when import items using overwrite to svn/git project

kjwang-talend:
TUP-14737 too slowly when import items using overwrite to svn/git project
https://jira.talendforge.org/browse/TUP-14737

### DIFF
--- a/main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/XmiResourceManager.java
+++ b/main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/XmiResourceManager.java
@@ -353,7 +353,7 @@ public class XmiResourceManager {
         return itemResource;
     }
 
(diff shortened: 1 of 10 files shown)


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/XmiResourceManager.java
```diff
@@ -353,7 +353,7 @@ private Resource createItemResource(boolean byteArrayResource, URI itemResourceU
         return itemResource;
     }
 
-    public void deleteResource(Resource resource) throws PersistenceException {
+    public void deleteResource(Resource resource, boolean isDeleteOnRemote) throws PersistenceException {
```
ggu-talend: The parameter is unused totally.

PR #808, opened 2016-12-20
TUP-16616 fix: Error "java.util.ConcurrentModificationException" when connecting to studio

cmeng-talend:
https://jira.talendforge.org/browse/TUP-16616

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/components/IComponentsFactory.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/components/IComponentsFactory.java
@@ -48,11 +48,7 @@ public interface IComponentsFactory {
 
     public Set<IComponent> getComponents();
 
-    public void addStdComponent(IComponent component);
-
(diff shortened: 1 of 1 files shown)


cmeng-talend:
related pull requests:
https://github.com/Talend/tcommon-studio-se/pull/808
https://github.com/Talend/tdi-studio-se/pull/883
https://github.com/Talend/tdi-studio-ee/pull/295
https://github.com/Talend/tesb-studio-ee/pull/39

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/components/IComponentsFactory.java
```diff
@@ -47,6 +48,26 @@
 
     public Set<IComponent> getComponents();
 
+    public void addStdComponent(IComponent component);
+
+    public void addStdComponents(Collection<IComponent> components);
```
ggu-talend: Maybe, no need add too much method like this. just one addXXX/removeXXX, should be enough.

PR #815, opened 2016-12-26
bugfix: TUP-6549  tFixFlowInput on the Ubuntu job:show an error when I

guol-talend:
click add button before add schema


main/plugins/org.talend.commons.ui/src/main/java/org/talend/commons/ui/swt/tableviewer/behavior/DefaultCellModifier.java
```diff
@@ -54,6 +54,9 @@ public DefaultCellModifier(TableViewerCreator tableViewerCreator) {
      * @see org.eclipse.jface.viewers.ICellModifier#canModify(java.lang.Object, java.lang.String)
      */
     public boolean canModify(Object bean, String idColumn) {
+        if (idColumn != null && idColumn.equals("__MASKED_COLUMN__")) {
```
ggu-talend: Don't think, it's good solution like this.

PR #826, opened 2017-01-09
bugfix: TUP-4619 Studio Cloudera cluster configuration issue

guol-talend:


main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/wizard/AbstractNamedWizardPage.java
```diff
@@ -75,7 +75,7 @@ protected static IStatus createStatus(int severity, String message) {
 
 
     protected void evaluateName(String name) {
-        if (name == null || name.length() == 0) {
+        if (name == null || name.length() == 0 || StringUtils.trimToNull(name) == null) {
             nameStatus = createStatus(IStatus.ERROR, Messages.getString("PropertiesWizardPage.NameEmptyError")); //$NON-NLS-1$
         } else if (name.startsWith(" ") //$NON-NLS-1$
```
ggu-talend: Maybe, also need check the '\t', in the begin of name or even middle of name.

main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/wizard/AbstractNamedWizardPage.java
```diff
@@ -75,7 +75,7 @@ protected static IStatus createStatus(int severity, String message) {
 
 
     protected void evaluateName(String name) {
-        if (name == null || name.length() == 0) {
```
ggu-talend: Better add one comment for the fix, what's fixed here, should be more clearly.

PR #845, opened 2017-01-22
TDI-37755: Add a common method to remove LineSeparator in MEMO.

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-37755

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
@@ -939,7 +939,7 @@ public class NodeUtil {
      */
     public static String replaceCRLFInMEMO_SQL(String original) {
  
-        return replaceLineSeparatorInMEMO(original,"(\n|\r\n)","\\n");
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
```diff
@@ -1071,4 +1071,49 @@ public static MetadataTalendTypeFilter createMetadataTalendTypeFilter(INode node
...
+    /**
+    *
+    * DOC Aim to resolve TDI-37755 ,just replace content in double quotation.
+    *
+    * @param original 
+    * @param regex: '\r' for Mac, '\n' for Linux/Unix, '\r\n' for Windows
+    * @param replacement: character to replace the line separator.
+    * @return
+    */
+   public static String replaceLineSeparatorInMEMO(String original,String regex, String replacement) {
```
jzhao-talend: I think you can call this method inside `replaceCRLFInMEMO_SQL(String original)` to reduce the duplicated code.
Should be `replaceLineSeparatorInMEMO(ElementParameterParser.getValue(node, "__MESSAGE__"),"(\r\n|\n)","\\n")`
You can try and test this.

PR #850, opened 2017-01-30
feat(TDI-38068): integrate snowflake component

chmyga:
* Add snowflake component to studio repository.


gonchar-ivan:
I've checked mvn clean package on
tcommon-studio-se/main/plugins/org.talend.designer.maven.tos/resources/components

It's ok. Added snowflake component itself and its dependencies

PR #854, opened 2017-02-08
Revert "TDI-37755: Add a common method to remove LineSeparator in MEMO."

pyzhou-talend:
Reverts Talend/tcommon-studio-se#845


pyzhou-talend:
"/r" is only used by old version of Mac. We can omit it and sothat the origin replaceCRLFInMEMO_SQL could be reserved.

wwang-talend:
only revert the commit, LGTM

PR #863, opened 2017-02-20
TUP-17003: Add Deployment tab in job setting view.

zwxue:


ggu-talend:
All commits are integrated in TUP-16719 task.

PR #888, opened 2017-03-06
TUP-17229:context value be overwrited after retrieve schema

wchen-talend:


nrousseau:
where are the junits...?

wchen-talend:
Need to use mock and power mock to do the test . So I didn't do the junit .

main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/model/ProjectNodeHelper.java
```diff
@@ -528,14 +528,11 @@ public static void fillCatalogAndSchemas(IMetadataConnection iMetadataConnection
...
-            if (((DatabaseConnection) iMetadataConnection.getCurrentConnection()).getUiSchema() != null) {
+            if (!temConnection.isContextMode()
+                    && ((DatabaseConnection) iMetadataConnection.getCurrentConnection()).getUiSchema() != null) {
                 temConnection.setUiSchema(((DatabaseConnection) iMetadataConnection.getCurrentConnection()).getUiSchema());
             }
 
-            if (((DatabaseConnection) iMetadataConnection.getCurrentConnection()).getSID() != null) {
-                temConnection.setSID(((DatabaseConnection) iMetadataConnection.getCurrentConnection()).getSID());
-            }
-
```
hcyi: no need set the SID ?

wchen-talend: SID already set in the dbInstance.fillUIConnParams(iMetadataConnection, temConnection) if not context mode

PR #914, opened 2017-03-20
feat(TBD-4707): Add Google Dataproc distribution and its context support

pbailly:
Google dataproc need to have the contexts  files integrated to the jars.
This PR will allow the code generator to do that.

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
@@ -65,9 +65,9 @@ import org.talend.repository.model.IProxyRepositoryService;
 @SuppressWarnings("unchecked")
 public final class ProcessUtils {
 
-    private static boolean               jarWithContext;
+    private static boolean jarWithContext;
 
-    private static List<IProcess>        fakeProcesses        = new ArrayList<IProcess>();
+    private static List<IProcess> fakeProcesses = new ArrayList<IProcess>();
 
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
```diff
@@ -961,20 +962,22 @@ public static boolean isRequiredPigUDFs(IProcess process) {
         return false;
     }
 
-    public static boolean isHDInsight() {
-        return isHD;
+    public static boolean jarNeedToContainsContext() {
```
RyanSkraba: English: `jarNeedsToContainContext()` -- but purely optional.

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
```diff
@@ -961,20 +962,22 @@ public static boolean isRequiredPigUDFs(IProcess process) {
...
 
-    public static void setHDInsight(boolean isHD) {
-        ProcessUtils.isHD = isHD;
+    public static void setJarWithContext(boolean jarWithContext) {
+        ProcessUtils.jarWithContext = jarWithContext;
     }
 
-    public static boolean isDistributionExist(ProcessItem processItem) {
+    /** Find the distribution where the generated jar rquired to have the context files inside **/
+    public static boolean needToHaveContextInsideJar(ProcessItem processItem) {
```
RyanSkraba: `needsToHaveContextInsideJar`

PR #920, opened 2017-03-28
TESB-18908:Memory-Leak in TalendDataSource implementation

csuntalendbj:
There are still some components are using the close method, so we could remove the close method later.


jlolling:
Great! Thanks.

PR #923, opened 2017-03-30
TBD-4738: Add db mapping for BigQuery spark components

bchen-talend:


main/plugins/org.talend.core.runtime/mappings/mapping_BigQuery.xml
```diff
@@ -0,0 +1,104 @@
+<?xml version="1.0"?>
```
amarouni: Where is this file used ?

main/plugins/org.talend.core.runtime/mappings/mapping_BigQuery.xml
```diff
@@ -0,0 +1,104 @@
...
+                    <dbType type="INTEGER" default="true" />
+                </talendType>
+                <talendType type="id_Object">
+                </talendType>    
+                <talendType type="id_Short">
+                </talendType>
+                <talendType type="id_String">
+                    <dbType type="STRING" default="true" />
+                    <dbType type="DATETIME" />
+                    <dbType type="DATE" />
```
amarouni: Why is the `DATE` mapped to `String` ? Why cannot we map it to `Date` (`java.util.Date`) ?

main/plugins/org.talend.core.runtime/mappings/mapping_BigQuery.xml
bchen-talend: For tBigQueryOutput component, which need to build bigquery table schema for insert & create table, also for writing jsonObject to insert.

main/plugins/org.talend.core.runtime/mappings/mapping_BigQuery.xml
bchen-talend: As BigQuery's date type support subsecond, https://cloud.google.com/bigquery/docs/reference/standard-sql/data-types#datetime-type, so use String for the moment.

PR #929, opened 2017-04-05
DEVOPS-1222 Pull MDM Webservice stub artifacts from Nexus during Stud…

wsliu-talend:
…io build


wsliu-talend:
The goal of the implementation is to pull webservice.jar file from zl nexus server instead of putting jar file to lib folder by hand everytime.

sbliu:
this commit is not valid anymore.

wsliu-talend:
Implemented by Dev in other PRs.

PR #940, opened 2017-04-11
TMDM-10190: Extract webservice jar to fragment and change the jar name.

sbliu:


main/plugins/org.talend.libraries.mdm.webservice.ce/build.properties
```diff
@@ -1,2 +1,3 @@
 bin.includes = META-INF/,\
-               lib/
+               lib/,\
+               lib/mdm_webservice_tomcat.jar
```
nrousseau: small thing, but it's no use to have lib/ + lib/mdm_webservice_tomcat.jar

PR #959, opened 2017-04-19
TDQ-13437 make tmultipattern support special cell editor

yyin-talend:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/update/RepositoryUpdateManager.java
```diff
@@ -1270,6 +1272,17 @@ protected void checkAndSetParameters(IProcess2 process2) {
                         isAddColumn = false;
                     }
                 }
+                
+                //Added TDQ-13437: when the job contains tMultiPattern, and update the REPOSITORY_PROPERTY_TYPE value to the current modified pattern id.
+                for(INode node: process2.getGraphicalNodes()){
+                    if(node.getElementName().startsWith("tMultiPattern") && parameter instanceof Item){
```
nrousseau: We should really avoid the check of fixed component here.
Better review if possible with ggu again later

ggu-talend: maybe try "node.getComponent().getName()" better.

PR #963, opened 2017-04-20
TDQ-13464: fix running the match analysis with "Store on disk" option

msjian:
get error


main/plugins/org.talend.libraries.persist.lookup/META-INF/MANIFEST.MF
```diff
@@ -18,3 +18,5 @@ Export-Package: org.talend.commons.utils.data.map,
  org.talend.designer.components.thash.io.hashimpl
 Bundle-Vendor: .Talend SA.
 Bundle-ActivationPolicy: lazy
+Require-Bundle: org.talend.librariesmanager,
```
ggu-talend: Why required the non-lib bundle?

msjian: i do rename org.talend.libraries.persist.lookup to org.talend.persist.lookup

PR #964, opened 2017-04-20
TESB-17306:Connect Talend (ESB) Studio to Nexus

csuntalendbj:
Pls check and merge(Feature)

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/EParameterFieldType.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/EParameterFieldType.java
@@ -109,6 +109,8 @@ public enum EParameterFieldType {
 
     GROUPING_CAMPAIGN_CHOOSER,// htyin added for DQ Matching components
     MULTI_PATTERN,//yyin added TDQ-13437
+    BUTTON_SYNC,
+    BUTTON_CHECK,
     ;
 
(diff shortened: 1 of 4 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/EParameterFieldType.java
```diff
@@ -109,6 +109,8 @@
 
     GROUPING_CAMPAIGN_CHOOSER,// htyin added for DQ Matching components
     MULTI_PATTERN,//yyin added TDQ-13437
+    BUTTON_SYNC,
+    BUTTON_CHECK,
```
nrousseau: can't you have some more explicit controller name...?

csuntalendbj: Actually , I would like to reuse the "Button" type,But it can not handle the call back event, so what about BUTTON_SYNC_LIBS and BUTTON_CHECK_LIBS?

csuntalendbj: My related controllers names are ButtonSyncController and ButtonCheckController , you mean I should specify concrete name, e.g., CConfigNexusSyncButtonController  and CConfigNexusCheckButtonController?

PR #974, opened 2017-04-24
TDI-38627:NUMERIC format problem on tAS400Output

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-38627


pyzhou-talend:
I have checked all the existing types.

PR #1003, opened 2017-05-03
fix(TUP-17583): The default java version is 1.7 for new created remote project.

zwxue:


dlin-talend:
it exist long time ago ,it's  a real problem here

PR #1026, opened 2017-05-15
fix(TUP-17643): Locally checked out Branch does not contains the project settings from the master -GIT.

zwxue:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/runtime/projectsetting/ProjectPreferenceManager.java
```diff
@@ -160,9 +173,29 @@ public boolean getBoolean(String key) {
...
      */
     public void save() {
-        if (service != null) {
-            service.storeProjectPreferences(getPreferenceStore());
+        if (runProcessService != null) {
+            runProcessService.storeProjectPreferences(getPreferenceStore());
         }
     }
+    
+    public void reload() {
```
ggu-talend: Seems there are 2 main problems:
- always use current project, even for some reference project?
- do addPropertyChangeListener in init method, so means, when reload, will add new reload listener again and again? (maybe should remove the old one first?) 
- additionally, if reload, need remove the inner listener for ScopedPreferenceStore or not, maybe no need.

PR #1035, opened 2017-05-21
fix(TUP-17746): fix search artifacts from nexus

nrousseau:


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/maven/ShareLibrareisHelper.java
```diff
@@ -142,8 +142,6 @@ public IStatus shareLibs(Job job, IProgressMonitor monitor) {
                         for (String groupId : groupIds) {
                             searchResults.addAll(instance.search(customServer.getServer(), customServer.getUserName(),
                                     customServer.getPassword(), customServer.getRepositoryId(), groupId, null, null));
-                            searchResults.addAll(instance.search(customServer.getServer(), customServer.getUserName(),
-                                    customServer.getPassword(), customServer.getSnapshotRepId(), groupId, null, null));
```
wchen-talend: why remove the search from snapshot repository ?

PR #1045, opened 2017-05-26
fix(TUP-17897): talend metadata fetch the xml struct wrongly or wordy when namespace exist

wwang-talend:


main/plugins/org.talend.datatools.xml/src/main/java/org/talend/datatools/xml/utils/SchemaPopulationUtil.java
```diff
@@ -393,30 +392,23 @@ private void insertNode(String treatedPath, String prefix, String uri) {
...
                         try {
+                            String prefix = namespace.getKey();
                             namespaceNode.setDataType(prefix);
                         } catch (OdaException e) {
                             // nothing
                         }
                         namespaceNode.setType(ATreeNode.NAMESPACE_TYPE);
+                        String uri = namespace.getValue();
                         namespaceNode.setValue(uri);
-                        ((ATreeNode) root.getChildren()[0]).addAsFirstChild(namespaceNode);
```
wwang-talend: it's not right to add the namespace to root element, should use the element which declared the namespace

PR #1048, opened 2017-05-30
fix(TDI-34726): Implement SAX filter for removing of namespaces

iistomin-gl:
* Introduced SAX filter which removes namespace mappings from XML

This PR is part of https://github.com/Talend/tdi-studio-se/pull/1319

### DIFF
--- a/main/plugins/org.talend.librariesmanager/resources/java/routines/system/xml/sax/NamespaceFilter.java
+++ b/main/plugins/org.talend.librariesmanager/resources/java/routines/system/xml/sax/NamespaceFilter.java
@@ -1,0 +1,102 @@
+package routines.system.xml.sax;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Stack;
+
+import org.xml.sax.Attributes;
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.librariesmanager/resources/java/routines/system/xml/sax/NamespaceFilter.java
```diff
@@ -0,0 +1,102 @@
...
+import org.xml.sax.helpers.XMLFilterImpl;
+
+/**
+ *
+ */
+public class NamespaceFilter extends XMLFilterImpl {
+
+    private Matcher matcher;
+
+    private Stack<Context> nsStack = new Stack<Context>();
```
dmytro-sylaiev: Why you didn't use ArrayDeque here?

iistomin-gl: Well, you are right. Stack is too old and based on Vector (which is synchronized).

PR #1057, opened 2017-06-02
fix(TUP-17328):Commandline is not able to get new jars in the lib folder

hwang-talend:


main/plugins/org.talend.librariesmanager.ui/src/main/java/org/talend/librariesmanager/utils/DownloadModuleRunnable.java
```diff
@@ -127,6 +129,9 @@ private void downLoad(final IProgressMonitor monitor) {
             ILibraryManagerService libraryManagerService = (ILibraryManagerService) GlobalServiceRegister.getDefault()
                     .getService(ILibraryManagerService.class);
             libraryManagerService.forceListUpdate();
+
+            ShareLibsOnStartup shareLib = new ShareLibsOnStartup();
```
wchen-talend: ShareLibsOnStartup will check all modules .
After download one jar , no need to check all modules.

PR #1059, opened 2017-06-05
fix(TDI-32437): fix the Bug_TDI_29037_tOracleInput_65535 tuj

wwang-talend:


main/plugins/org.talend.librariesmanager/resources/java/routines/system/JDBCUtil.java
```diff
@@ -65,4 +65,22 @@ public static Date getDate(ResultSet rs, int index) throws java.sql.SQLException
         return null;
     }
     
+    //decrease the get method call number
+    public static Double getDoubleObject2(ResultSet rs, int index) throws java.sql.SQLException {
```
pyzhou-talend: I think this method can have a better name which describe its function precisely,like getNonzeroDoubleObject().  but it's trivial thing.

wwang-talend: The result can be zero.

pyzhou-talend: ok. forgot that then

PR #1074, opened 2017-06-09
fix(TCOMP-291):add method to find dynamic property

chmyga:
Related to https://jira.talendforge.org/browse/TCOMP-291
and is required for https://github.com/Talend/tdi-studio-se/pull/1366

TO BE MERGED AFTER 6.4.1 RELEASE


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/ContextParameterUtils.java
```diff
@@ -248,14 +248,18 @@ public static String getNewScriptCode(final String name, ECodeLanguage language)
      * only for new script code and the first variables. and if there is no variable in code, return null.
      */
     public static String getVariableFromCode(String code) {
+        return getVariableFromCode(code, JAVA_NEW_CONTEXT_PREFIX);
+    }
+
+    public static String getVariableFromCode(String code, String prefix) {
```
gonchar-ivan: I would suggest to make it private... however, it's not a problem

PR #1079, opened 2017-06-09
fix(TUP-17773): avoid checking the non-xml properties files

ggu-talend:


main/plugins/org.talend.repository.view/src/org/talend/repository/navigator/RepoViewCommonViewer.java
```diff
@@ -419,8 +425,31 @@ public void run() {
                     if (xrm.isPropertyFile(fileUpdated)) {
                         IFile file = ResourcesPlugin.getWorkspace().getRoot().getFile(new Path(fileUpdated));
                         if (file != null && file.isAccessible()) {
```
nrousseau: seems like a big problem of performance possible if we must recheck all the time.
Maybe no need this no? since we know already which folder need a refresh..

I think simply the components folder should never trigger the event "Constant.FILE_LIST_EVENT_PROPERTY_KEY", then it will not even call this code

PR #1081, opened 2017-06-09
fix(TBD-5163): Remove the dependencies signature in the fat jar

rdubois:


test/plugins/org.talend.utils.test/src/org/talend/utils/files/FileUtilsTest.java
```diff
@@ -165,4 +163,26 @@ public void getAllFilesFromFolder3() throws IOException {
...
+    @Test
+    public void deleteFilesTest() throws IOException {
+
+        folder1.newFile("myfile.txt"); //$NON-NLS-1$
+        folder1.newFile("myfile2.txt"); //$NON-NLS-1$
+        folder1.newFile("myfile2.zip"); //$NON-NLS-1$
+
+        File folder1File = folder1.getRoot();
+
+        FileUtils.deleteFiles(folder1File, name -> false);
```
sgandon: You may use () -> false here I think.

test/plugins/org.talend.utils.test/src/org/talend/utils/files/FileUtilsTest.java
```diff
@@ -165,4 +163,26 @@ public void getAllFilesFromFolder3() throws IOException {
...
+        assertEquals(folder1File.listFiles().length, 3);
+
+        FileUtils.deleteFiles(folder1File, name -> name.endsWith(".txt")); //$NON-NLS-1$
+        assertEquals(folder1File.listFiles().length, 1);
```
sgandon: I would suggest that you also check this is the right file (myfile2.zip) that is left.

test/plugins/org.talend.utils.test/src/org/talend/utils/files/FileUtilsTest.java
rdubois: It wouldn't compile.

test/plugins/org.talend.utils.test/src/org/talend/utils/files/FileUtilsTest.java
rdubois: +1. Done

PR #1083, opened 2017-06-12
fix(TUP-18022):CI builder: item run fail on LogoffProjectCommand.

zwxue:


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
```diff
@@ -2026,10 +2026,12 @@ public void logOnProject(Project project, IProgressMonitor monitor) throws Login
     public void logOffProject() {
         // getRepositoryContext().setProject(null);
         repositoryFactoryFromProvider.logOffProject();
-        ProjectRepositoryNode root = ProjectRepositoryNode.getInstance();
-        if (root != null) {
-            root.setEnableDisposed(true);
-            root.dispose();
+        if (CommonsPlugin.isHeadless()) {
```
nrousseau: should be: if (!CommonsPlugin.isHeadless()) {


We want to avoid this code if it's commandline.. not execute it if it's commandline

PR #1104, opened 2017-06-15
fix(TUP-18073): enhance and unify the component folders

ggu-talend:


ggu-talend:
will create on master first

PR #1117, opened 2017-06-22
fix(TUP-6860)Access has Error when check connection with password

Seloen:
https://jira.talendforge.org/browse/TUP-6860


main/plugins/org.talend.repository.metadata/src/main/java/org/talend/repository/metadata/i18n/messages.properties
```diff
@@ -879,7 +879,7 @@ DatabaseForm.SystemSQL=System SQL Statement
 DatabaseForm.check=&Check
 DatabaseForm.checkConnectionTitle=Check Connection 
 DatabaseForm.checkFailure=Connection failure.
-DatabaseForm.checkFailureTip=You must change the Database Settings.
```
ggu-talend: No need change the messages, I think

PR #1137, opened 2017-07-03
feat(TCOMP-564): add new code parts

chmyga:
Related PRs are:

https://github.com/Talend/tdi-studio-se/pull/1477
https://github.com/Talend/components/pull/872

Related task:
https://jira.talendforge.org/browse/TCOMP-564

This PR should be merged second after https://github.com/Talend/components/pull/872


gonchar-ivan:
However, we should be sure if it is not used in switch clauses or something like that, because otherwise this change may break switch

undx:
Done in PRs :
- Talend/components#1107
- Talend/tdi-studio-se#1862
- Talend/tcommon-studio-se#1350
Closing.

PR #1144, opened 2017-07-07
fix(TUP-5360): Retrieve Salesforce Modules option is missing

ycbai:


main/plugins/org.talend.repository.view/src/org/talend/repository/viewer/action/RepoDoubleClickAction.java
```diff
@@ -415,6 +417,18 @@ private ITreeContextualAction getAction(RepositoryNode obj) {
...
         }
+
+        IGenericWizardService wizardService = null;
+        if (GlobalServiceRegister.getDefault().isServiceRegistered(IGenericWizardService.class)) {
+            wizardService = (IGenericWizardService) GlobalServiceRegister.getDefault().getService(IGenericWizardService.class);
+        }
+        if (wizardService != null) {
+            ITreeContextualAction defaultAction = wizardService.getDefaultAction(obj);
+            if (defaultAction != null) {
+                return defaultAction;
```
gonchar-ivan: if (defaultAction != null) then return it
if (defaultAction == null) then .... return null

You don't need `if (defaultAction != null)` condition.
You may just leave `return defaultAction`
The result will be the same

nrousseau:
to close, handled at the same time as the jdbc feature

PR #1151, opened 2017-07-12
TUP-18212:Hadoop Property should not sync to HBase parameters in Spark job

hcyi:


ycbai:
This is not a bug, so no need to change anything.

PR #1161, opened 2017-07-19
kjwang/fix(TUP-18276):PATCH broke Build Job with sub-jobs from Reference

kjwang-talend:
kjwang/fix(TUP-18276):PATCH broke Build Job with sub-jobs from Reference project
https://jira.talendforge.org/browse/TUP-18276


ggu-talend:
Need test some cases:
  - only build the jobs which are from reference projects.
  - build the main jobs with reference jobs.

Maybe need change the pom and assembly template to adapt for jobs from ref-projects.

kjwang-talend:
I have checked, build main job and main job with reference jobs is work well. There is a problem that will lost routines.jar when build a reference jobs. zwxue is checking now.

PR #1186, opened 2017-08-08
TESB-20067 Patched Axis 1.4 library is added to Axis dependencies bundle

apoltavtsev:


apoltavtsev:
Name of patched Axis library is changed from axis-p1-6.0.0.jar to axis-p1.jar

PR #1188, opened 2017-08-10
feat TUP-17207:Automate configuration of Project Configuration

hcyi:


hcyi:
After check with ggu , we no need this .

PR #1201, opened 2017-08-24
fix(TUP-18388): add some new junits to enhance and fix the failed ones.

hcyi:


test/plugins/org.talend.core.runtime.test/src/org/talend/core/model/utils/ParameterValueUtilTest.java
```diff
@@ -467,10 +467,27 @@ public void testRenameValues4SQLAndGlobleMap() {
         expectRetValue = "\"select A.id, A.name form \"+context.table+\" A where A.name= \"+((String)globalMap.get(\"tFileList_2_CURRENT_FILE\"))";
         // if flag is false, means is from SQL. but when replace the globlemap, will be problem.
         retValue = ParameterValueUtil.renameValues(testString, "tFileList_1", "tFileList_2", false);
-        Assert.assertTrue("testRenameValues4SQLAndGlobleMap", expectRetValue.equals(retValue));
+        Assert.assertFalse("testRenameValues4SQLAndGlobleMap", expectRetValue.equals(retValue));
```
wchen-talend: The original junit is necessary , as I understand the ((String)globalMap.get(\"tFileList_1_CURRENT_FILE\")) should be replaced . otherwise it means the copy&past won't replace for global map like this and the joblet can't use global map at all . Please have a check with this case.

ggu-talend: Yes, maybe just check startWith for unique name, ingore others.

PR #1219, opened 2017-09-06
TESB-20303 - Ensure reference projcet recognition for project names w…

andreasmattes:
…ith spaces.

This fix ensures that the reference project is recognized by its name even if label and technical label differ by more than just upper/lower case due to characters like spaces in the project label which are escaped in the technical label.

While the class containing the code change is commonly used, the changed code itself is only invoked for routes with routelets. Therefore, adverse side effects are very unlikely for other clients of the modified class.


nrousseau:
would be better to check why we need a specific code only for routelets.. while we totally don't need for other kinds of items it seems. (or maybe bug was not found for other items?)
Fix might work, but it's just try to fix some code that was already wrong before.

main/plugins/org.talend.core.runtime/src/main/java/org/talend/repository/ProjectManager.java
```diff
@@ -1,6 +1,6 @@
 // ============================================================================
 //
-// Copyright (C) 2006-2017 Talend Inc. - www.talend.com
+// Copyright (C) 2006-2016 Talend Inc. - www.talend.com
```
yanbasic: The copyright year is updated to 2017

andreasmattes:
All code specific for routelet handling has been removed from class ProjectManager and factored out into the new class RouteUtil in package in package org.talend.core.model.utils - the new class also belongs to bundle org.talend.core.runtime so that visibility during code generation is not affected.

PR #1262, opened 2017-10-27
feat(TUP-18316): find a way to lower the cost of adding/upgrading BD Distros: Wizard (Import conf / build conf)

cmeng-talend:
https://jira.talendforge.org/browse/TUP-18316


cmeng-talend:
already merged in https://github.com/Talend/tcommon-studio-se/pull/1357

PR #1271, opened 2017-11-03
fix(TUP-18442):Enforce maven coordinate resolution consistency

wchen-talend:


test/plugins/org.talend.designer.maven.test/src/main/java/org/talend/designer/maven/utils/PomUtilTest.java
```diff
@@ -595,4 +595,51 @@ public void test_getJobVersionForPomProperty_arguments_property() {
...
+
+    }
+
+    @Test
+    public void getAbsArtifactPath() {
+        MavenArtifact artifact = new MavenArtifact();
+        artifact.setGroupId("org.talend.libraries");
+        artifact.setArtifactId("getAbsArtifactPath");
+        artifact.setVersion("1.0.0");
+        String absArtifactPath = PomUtil.getAbsArtifactPath(artifact);
```
ggu-talend: I'm not sure does it work on JUnit server( with docker). 
if in .m2, not existed, will failure.

ggu-talend: seems we provide the "commons-lang3" by default in m2. so should be ok

PR #1303, opened 2017-11-30
feat(TBD-5415) Add GCS User Agent String in Constant class

yu-talend:


main/plugins/org.talend.librariesmanager/resources/java/routines/system/Constant.java
```diff
@@ -8,14 +8,20 @@
  */
 public abstract class Constant {
 
+    private static final String version_studio = org.talend.commons.utils.VersionUtils.getDisplayVersion();
```
lbourgeois: Shoud be TALEND_STUDIO_VERSION to be consistent with others constants

yu-talend:
Hello,

I'm sorry and in fact there are some problems for the previous changes.

It doesn't work because the org.talend.commons.utils.VersionUtils class is in fact in code generating (eclipse) level (or rather to say it is not packaged and included in the Studio classpath ?). So as a consequence we don't have it in the generated project in Studio, so I moved it (version retrieving) to javajet level as shown in the 3 other related commits
https://github.com/Talend/tdi-studio-se/pull/1749
https://github.com/Talend/tbd-studio-se/pull/595
https://github.com/Talend/tbd-studio-ee/pull/388
to let us get the version.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/Constant.java
```diff
@@ -6,16 +6,24 @@
...
      */
     public static final String dateDefaultPattern = "dd-MM-yyyy";
 
     /**
-     * the default user agent string for AWS components
+     * the default user agent string for AWS and Azure components
      */
-    public static final String TALEND_USER_AGENT = "APN/1.0 Talend/6.4 Studio/6.4";
-
+    public static String getUserAgent(String version_studio) {
```
lbourgeois: Please rename _version_studio_ to _studioVersion_

PR #1314, opened 2017-12-08
TUP-18826 Need a notification mecanism when a connection between

kjwang-talend:
TUP-18826 Need a notification mecanism when a connection between components is added/removed
https://jira.talendforge.org/browse/TUP-18826


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/AbstractNode.java
```diff
@@ -999,8 +1000,16 @@ public void removeInput(IConnection connection) {
...
+        Iterator<IMetadataTable> it = this.getMetadataList().iterator();
+        boolean found = false;
+        while (it.hasNext() && !found) {
+            IMetadataTable table = it.next();
+            if (table.getTableName().equals(connection.getMetaName())) {
+                it.remove();
+                found = true;
+            }
+        }
+        getProcess().removeUniqueConnectionName(connection.getName());
```
nrousseau: This is an example to provide to tdm team in the jira.
It should not be commited like this, the function should be empty still

PR #1316, opened 2017-12-08
task(TCOMP-691): create context menu for Tacokit node in repository panel

cmeng-talend:
https://jira.talendforge.org/browse/TCOMP-691

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/repository/ERepositoryObjectType.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/repository/ERepositoryObjectType.java
@@ -556,8 +556,6 @@ public class ERepositoryObjectType extends DynaEnum<ERepositoryObjectType> {
     public final static ERepositoryObjectType METADATA_SAP_BW_INFOOBJECT = ERepositoryObjectType
             .valueOf("METADATA_SAP_BW_INFOOBJECT"); //$NON-NLS-1$
 
-    public final static ERepositoryObjectType METADATA_TACOKIT = ERepositoryObjectType.valueOf("TACOKIT");
-
     private static Map<String, ERepositoryObjectType> typeCacheById = new HashMap<String, ERepositoryObjectType>();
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/repository/ERepositoryObjectType.java
```diff
@@ -556,6 +556,8 @@
     public final static ERepositoryObjectType METADATA_SAP_BW_INFOOBJECT = ERepositoryObjectType
             .valueOf("METADATA_SAP_BW_INFOOBJECT"); //$NON-NLS-1$
 
+    public final static ERepositoryObjectType METADATA_TACOKIT = ERepositoryObjectType.valueOf("TACOKIT");
```
nrousseau: why is it needed? we don't need to put nodes from extension point here.

main/plugins/org.talend.repository.localprovider/src/main/java/org/talend/repository/localprovider/model/LocalRepositoryFactory.java
```diff
@@ -2802,6 +2815,21 @@ public void create(Project project, Item item, IPath path, boolean... isImportIt
...
                 itemResource = create(project2, (EDIFACTConnectionItem) item, ERepositoryObjectType.METADATA_EDIFACT, path);
                 break;
+            case PropertiesPackage.CONNECTION_ITEM:
+                // connection item may be used by extention point
+                final int classifierID = eClass.getClassifierID();
+                boolean created = false;
+                for (IRepositoryContentHandler handler : RepositoryContentManager.getHandlers()) {
+                    itemResource = handler.create(project2, item, classifierID, path);
+                    if (itemResource != null) {
+                        screenshotsResource = handler.createScreenShotResource(project2, item, classifierID, path);
```
nrousseau: there is really screenshots for connections...? normally it's only for jobs

PR #1321, opened 2017-12-13
fix(TDI-39791): extract JSONArray from JSONObject routines

pteyssier:


pyzhou-talend:
This change looks good to me. But I don't have the authority to approve this PR.
FYI: I find that we have a JSONArray.java with almost the same content in org.talend.utils project. Since it cannot be used by customer, I think it doesn't affect this fix.

PR #1322, opened 2017-12-14
fix(TUP-19089):Unable to select or deselect used variables in select

hwang-talend:
context variables window


main/plugins/org.talend.core.ui/src/main/java/org/talend/core/ui/context/SelectRepositoryContextDialog.java
```diff
@@ -331,7 +331,8 @@ private void updateSubCheckedState(Object obj, boolean checked, boolean showWarn
             ContextParameterType param = (ContextParameterType) obj;
             treeViewer.setGrayed(param, false);
             if (helper.existParameterForJob(param)) {
-                treeViewer.setChecked(param, true); // checked always
+                //lead to TUP-19089
+                //treeViewer.setChecked(param, true); // checked always
```
nrousseau: Seems not too logic when see the code below.
Why do we have a if with nothing inside, any reason? what's the code after / which case does it handle?

PR #1334, opened 2017-12-20
fix(TDI-39697): The "Use custom object" in module list view doesn't work well for tjdbcoutput as it require the custom value must exists in the list above

wwang-talend:


wwang-talend:
before, for StringProperty with possible values will go in to this code block : 
{code}
if (GenericTypeUtils.isEnumType(property)) {
            List<?> possibleValues = property.getPossibleValues();
            if (possibleValues != null) {
                Object firstValue = null;
                if (!possibleValues.isEmpty()) {
                    firstValue = possibleValues.get(0);
                }
                String stringStoredValue = TalendQuoteUtils.removeQuotes(stringValue);
                for (Object possibleValue : possibleValues) {
                    if (possibleValue.toString().equals(stringStoredValue)) {
                        return possibleValue;
                    }
                }
                if (firstValue != null) {
                    return firstValue;
                }
            }
        }
{code}
So will return the first value before.

PR #1336, opened 2017-12-22
fix(TUP-19163): fix the failed junits

ggu-talend:


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
```diff
@@ -386,14 +384,32 @@ private boolean checkIfHaveDuplicateName(Project project, Item item, IPath path)
...
+                            MessageBox box = new MessageBox(currentShell, SWT.ICON_WARNING | SWT.OK | SWT.CANCEL);
+                            box.setText(Messages.getString("ProxyRepositoryFactory.JobNameErroe")); //$NON-NLS-1$
+                            box.setMessage(Messages.getString("ProxyRepositoryFactory.Label") + " " + name + " " + Messages.getString("ProxyRepositoryFactory.ReplaceJob")); //$NON-NLS-1$ //$NON-NLS-2$//$NON-NLS-3$//$NON-NLS-4$
+
+                            if (box.open() == SWT.OK) {
+
+                                ok[0] = true;
+                            }
+                        }
+                    });
```
nrousseau: not right way to do here maybe no?
cf the function which call this one, there is normally this test:
if (isImportItem.length == 0) {

Means maybe junit need to setup the import flag

PR #1359, opened 2018-01-04
feat(TUP-19046): add signature on items

ggu-talend:


main/plugins/org.talend.commons.runtime/src/org/talend/commons/runtime/xml/XMLFileUtil.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * 
+ * DOC ggu class global comment. Detailled comment
+ */
+public final class XMLFileUtil {
+
+    public static Document loadDoc(InputStream stream) throws ParserConfigurationException, SAXException, IOException {
+        try {
+            DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
```
coheigea: DocumentBuilderFactory is expensive to create so this class could be private static final in this Utility class. Also in a static block you must set the secure processing feature to true on it:

https://docs.oracle.com/javase/7/docs/api/javax/xml/XMLConstants.html#FEATURE_SECURE_PROCESSING

main/plugins/org.talend.commons.runtime/src/org/talend/commons/runtime/xml/XMLFileUtil.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    public static Document loadDoc(File file) throws ParserConfigurationException, SAXException, IOException {
+        return loadDoc(new BufferedInputStream(new FileInputStream(file)));
+    }
+
+    public static void saveDoc(Document doc, OutputStream output) throws TransformerException {
+        try {
+            TransformerFactory tf = TransformerFactory.newInstance();
```
coheigea: TransformerFactory could also be static final + should have secure processing enabled:

https://docs.oracle.com/javase/7/docs/api/javax/xml/XMLConstants.html#FEATURE_SECURE_PROCESSING

main/plugins/org.talend.libraries.apache.xml/pom.xml
```diff
@@ -1,12 +1,61 @@
...
+								<artifactItem>
+									<groupId>org.apache.xmlrpc</groupId>
+									<artifactId>xmlrpc-common</artifactId>
+									<version>${xmlrpc.version}</version>
+								</artifactItem>
+								<artifactItem>
+									<groupId>org.apache.xmlrpc</groupId>
+									<artifactId>xmlrpc-client</artifactId>
+									<version>${xmlrpc.version}</version>
+								</artifactItem>
```
coheigea: Why are xmlschema-core + xmlrpc included as dependencies here?

main/plugins/org.talend.commons.runtime/src/org/talend/commons/runtime/xml/XMLFileUtil.java
kjwang-talend: Modified follow your suggestion, thanks

main/plugins/org.talend.commons.runtime/src/org/talend/commons/runtime/xml/XMLFileUtil.java
kjwang-talend: Modified follow your suggestion, thanks

main/plugins/org.talend.libraries.apache.xml/pom.xml
ggu-talend: just do it same as other bundles without jar in github, when build, will copy it to bundle.

PR #1372, opened 2018-01-12
fix(TDI-35418): Change mapping for REAL numbers

Voievidko:
https://jira.talendforge.org/browse/TDI-35418

### DIFF
--- a/main/plugins/org.talend.core.runtime/mappings/mapping_Oracle.xml
+++ b/main/plugins/org.talend.core.runtime/mappings/mapping_Oracle.xml
@@ -68,6 +68,7 @@
 				</talendType>
 				<talendType type="id_Float">
 					<dbType type="FLOAT" default="true"/>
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.core.runtime/mappings/mapping_Oracle.xml
```diff
@@ -75,7 +75,7 @@
 					<dbType type="INTEGER"/>
 					<dbType type="PLS_INTEGER"/>
 				</talendType>
-				<talendType type="id_Long">
+				<talendType type="id_Float">
```
gonchar-ivan: here, you shouldn't replace Long by Float, because you will have 2 entries for Float after.
I would say you need to move <dbType type="REAL"/> to correct Talend type. Also, please, check mapping for Long type. Maybe there is better type.

PR #1392, opened 2018-01-24
feat(TBD-6262): Generate ps1 file when building a spark job

zafkirTalend:


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
```diff
@@ -347,14 +353,11 @@ protected void addProperties(Model model) {
...
-        StringBuffer jvmArgsStr = new StringBuffer();
-        if (jvmArgs != null && jvmArgs.length > 0) {
-            for (String arg : jvmArgs) {
-                jvmArgsStr.append(arg);
-                jvmArgsStr.append(' ');
-            }
-        }
-        checkPomProperty(properties, "talend.job.jvmargs", ETalendMavenVariables.JobJvmArgs, jvmArgsStr.toString());
+        String jvmArgsStr = Arrays.stream(jvmArgs).collect(Collectors.joining(" "));
+        String jvmArgsStrPs1 = "\'" + Arrays.stream(jvmArgs).collect(Collectors.joining("\' \'")) + "\'";
```
kcoepeau: Java 8+ API may prevent compiling if some products (like < Studio 6.4) that are < Java 8 compliant.

kcoepeau: Why not using String.join("\' \'", Arrays.asList(jvmArgs)) ?

PR #1397, opened 2018-01-26
[TESB-20924] cConfig cannot install the external jar

vbogatyrov:
In scope of the commit:

- Prevent appearance of orange "Additional library required" bar with "Install..." button for cConfig component when exteran ljar is added to dependency list (ModuleListCellEditor.java)

- Added method to LibraryService API to publish jars only to local studio reposiotry, without attempting to reach Nexus

### DIFF
--- a/main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/AbstractLibrariesService.java
+++ b/main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/AbstractLibrariesService.java
@@ -136,7 +136,7 @@ public abstract class AbstractLibrariesService implements ILibrariesService {
         }
         final File sourceFile = new File(decode);
 
-        localLibraryManager.deploy(sourceFile.toURI(), mavenUri, updateNexusJar, null);
+        localLibraryManager.deploy(sourceFile.toURI(), mavenUri, updateNexusJar);
 
         refreshLocal(new String[] { sourceFile.getName() });
         if (refresh) {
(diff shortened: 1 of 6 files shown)


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/AbstractLibrariesService.java
```diff
@@ -132,13 +136,12 @@ public void deployLibrary(URL source, String mavenUri, boolean refresh) throws I
         }
         final File sourceFile = new File(decode);
 
-        localLibraryManager.deploy(sourceFile.toURI(), mavenUri);
+        localLibraryManager.deploy(sourceFile.toURI(), mavenUri, updateNexusJar, null);
```
yanbasic: 'remove null' change to 
localLibraryManager.deploy(sourceFile.toURI(), mavenUri, updateNexusJar); would be ok.

main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/LibrariesService.java
```diff
@@ -263,4 +263,8 @@ public void deployLibrary(URL source, String mavenUri, boolean refresh) throws I
         getLibrariesService().deployLibrary(source, mavenUri, refresh);
     }
 
+    public void deployLibrary(URL source, String mavenUri, boolean refresh, boolean updateNexusJar) throws IOException {
```
yanbasic: add  @Override

PR #1407, opened 2018-02-01
TDQ-14775: Only fill specified catalog/schema for JDBC connection

qiongli:


main/plugins/org.talend.metadata.managment/src/main/java/org/talend/metadata/managment/utils/MetadataConnectionUtils.java
```diff
@@ -901,61 +901,87 @@ public static String createTechnicalName(final String functionalName) {
     public static List<String> getPackageFilter(Connection connection, DatabaseMetaData dbMetaData, boolean isCatalog) {
```
ggu-talend: If only for DatabaseConnection, better replace the Connection to "DatabaseConnection" directly

PR #1409, opened 2018-02-02
fix(TUP-19270):Cannot connect from studio to TAC via proxy server with basic authentication

hcyi:


main/plugins/org.talend.commons.runtime/src/org/talend/commons/utils/network/NetworkUtil.java
```diff
@@ -78,6 +82,53 @@ public static Authenticator getDefaultAuthenticator() {
         return null;
     }
 
+    public static void loadAuthenticator() {
+        // get parameter from System.properties.
+        String proxySet = System.getProperty("http.proxySet"); //$NON-NLS-1$
+        if (Boolean.TRUE.toString().equals(proxySet)) {
```
ggu-talend: Not sure it's safe or not, if system properties, can use Boolean.getBoolean("http.proxySet") instead.

main/plugins/org.talend.commons.runtime/src/org/talend/commons/utils/network/NetworkUtil.java
```diff
@@ -78,6 +82,53 @@ public static Authenticator getDefaultAuthenticator() {
...
+    }
+
+    public static void updateSvnkitConfigureFile(String srcFilePath, String destFilePath) {
+        // SVNFileUtil getSystemApplicationDataPath C:\ProgramData\\Application Data
+        // Note:ProgramData:Starting with Windows 10,this setting can no longer be used in provisioning packages.
+        String osName = System.getProperty("os.name");//$NON-NLS-1$
+        String osNameLC = osName == null ? null : osName.toLowerCase();
+        boolean windows = osName != null && osNameLC.indexOf("windows") >= 0;//$NON-NLS-1$
+        String proxySet = System.getProperty("http.proxySet"); //$NON-NLS-1$
+        if (windows && Boolean.TRUE.toString().equals(proxySet)) {
```
ggu-talend: Same as above

PR #1422, opened 2018-02-07
fix(TUP-19329):Dyanmic: when Import two dynamic configuration, second one cann't check success.

hcyi:


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/AbstractLibrariesService.java
```diff
@@ -273,6 +273,7 @@ public void undeployLibrary(String jarName) throws IOException {
     public void resetModulesNeeded() {
         ModulesNeededProvider.reset();
         ModuleStatusProvider.reset();
+        ModulesNeededProvider.getModulesNeeded().clear();
```
wchen-talend: The ModulesNeededProvider.reset(); will clear all modules , no need reset twice.

PR #1432, opened 2018-02-11
TESB-20834: fixed NPE for microservice type missing

yanbasic:
Hi, please review my fix by adding microservice type.

Yi


yanbasic:
fix the tuj build error

PR #1509, opened 2018-03-12
xlwang/fix(TUP-19663)Studio too slow to start

xlwang-tlnd:


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/ExtensionModuleManager.java
```diff
@@ -78,7 +79,14 @@
 
     public final static String DEFAULT_LIB_FOLDER = "lib"; //$NON-NLS-1$
 
-    private static ExtensionModuleManager manager = new ExtensionModuleManager();
+
+    public final static IExtensionPointLimiter extensionPoint = new ExtensionPointLimiterImpl(EXT_ID, MODULE_GROUP_ELE);
```
nrousseau: quite bad / dirty to set this in the static....  it can't be moved to the getInstance at least ?
I'm sure we can have first initialization of the instance with all the basic things needed.

And why do we have this here only for groups? As i can see we have the same kind of code for the collectSingle no?

main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/ExtensionModuleManager.java
```diff
@@ -78,7 +79,14 @@
...
+
+    public final static List<IConfigurationElement> extension = ExtensionImplementationProvider.getInstanceV2(extensionPoint);
```
nrousseau: As we said, from the feature of BD, the extension points can be dynamic.. so we need a way to reset them. It should not be static. (but initialized first, and with a method available to reset it)

And i guess we need another cache like this for the collectSingle... since it's another list of extension points

Better if initialized at the first call of the collect if the cache was not reset before instead of have a static here..

cmeng-talend:
changes merged in https://github.com/Talend/tcommon-studio-se/pull/1523

PR #1530, opened 2018-03-16
Cmeng/bugfix/tup 19436 improve performance

cmeng-talend:


cmeng-talend:
https://jira.talendforge.org/browse/TUP-19436

PR #1553, opened 2018-03-26
fix(TUP-19759):Unhandled event loop exception issue when create file xml metadata

hcyi:


hcyi:
Will be fixed on TUP-19727 .

PR #1558, opened 2018-03-27
feat(TUP-19824):Create poms for joblet and update job content.

zwxue:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/designer/runprocess/IProcessor.java
```diff
@@ -294,13 +294,11 @@ public Process run(String[] optionsParam, int statisticsPort, int tracePort, IPr
...
 
-    Set<ModuleNeeded> getNeededModules();
-
-    Set<ModuleNeeded> getNeededModules(boolean withChildrenJobs);
+    Set<ModuleNeeded> getNeededModules(int options);
 
     Set<JobInfo> getBuildChildrenJobs();
 
-    Set<JobInfo> getBuildChildrenJobs(boolean firstChildOnly);
+    Set<JobInfo> getBuildFirstChildrenJobs();
```
gonchar-ivan: It is dangerous to remove methods from an interface.
More safe way is:
1. add new method `Set<JobInfo> getBuildFirstChildrenJobs();` it will be overloaded method
2. Mark old method as deprecated with description

PR #1559, opened 2018-03-27
fix(TESB-21651): No-op backupPomFile in PomUtil if project is null.

andreasmattes:
Assuming that PomUtil.backupPomFile(...) can be invoked in cases where no project is present, it will no longer throw an NPE but no-op and return.


andreasmattes:
Closed as obsolete (fixed in a different place).

PR #1560, opened 2018-03-28
feat(TUP-19782)Add filter on the pom synchronization

jding-tlnd:
Add a filter test method

### DIFF
--- a/test/plugins/org.talend.designer.maven.test/src/main/java/org/talend/designer/maven/tools/AggregatorPomsHelperTest.java
+++ b/test/plugins/org.talend.designer.maven.test/src/main/java/org/talend/designer/maven/tools/AggregatorPomsHelperTest.java
@@ -17,17 +17,33 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import java.util.List;
+
 import org.apache.maven.model.Model;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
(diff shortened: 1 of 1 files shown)


test/plugins/org.talend.designer.maven.test/src/main/java/org/talend/designer/maven/tools/AggregatorPomsHelperTest.java
```diff
@@ -62,6 +71,45 @@ public void testAddToAndRemoveFromParentModules() throws Exception {
...
+         IFolder jobFolder = pomsFolder.getFolder("jobs").getFolder("process").getFolder("job1");
+         if (!jobFolder.exists()) {
+             jobFolder.create(true, true, null);
+         }
+         IFile jobPom = jobFolder.getFile("pom.xml");
+         if(!jobPom.exists()) {
+        	 Model jobModel = new Model();
+        	 jobModel.setModelVersion("4.0.0"); 
+        	 jobModel.setGroupId("org.example.aa"); 
+        	 jobModel.setArtifactId(projectModel.getGroupId());
```
nrousseau: artifactId = projectmodel.getGroupId?

nrousseau: agreed anyway that it's not that important for the test... but if someone else review the junit later, seems strange still..

jding-tlnd: oh sorry, wrong write

PR #1562, opened 2018-03-28
fix(TUP-19839):tSnowflakeRow --> "Use PreparedStatement" ->'Parameter

hwang-talend:
Value' can't use context.


hcyi:
no need this since i do some change , need review the code .

PR #1577, opened 2018-03-30
fix(TUP-19636):import previous spark with test case job, set spark

zwxue:
version to another one, but pom dependency still need previous spark
version jar

### DIFF
--- a/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/ProcessorDependenciesManager.java
+++ b/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/ProcessorDependenciesManager.java
@@ -13,7 +13,9 @@
 package org.talend.designer.maven.tools;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/ProcessorDependenciesManager.java
```diff
@@ -191,22 +207,47 @@ public static boolean updateDependencies(IProgressMonitor progressMonitor, Model
...
-                ILibraryManagerService repositoryBundleService = (ILibraryManagerService) GlobalServiceRegister.getDefault()
-                        .getService(ILibraryManagerService.class);
-                repositoryBundleService.installModules(neededLibraries, null);
+            for (ModuleNeeded moduleNeeded : neededLibraries) {
+                if ("junit.jar".equals(moduleNeeded.getModuleName())) { //$NON-NLS-1$
+                    if (GlobalServiceRegister.getDefault().isServiceRegistered(ILibraryManagerService.class)) {
+                        ILibraryManagerService repositoryBundleService = (ILibraryManagerService) GlobalServiceRegister
+                                .getDefault()
+                                .getService(ILibraryManagerService.class);
+                        repositoryBundleService.installModules(Arrays.asList(moduleNeeded), null);
```
nrousseau: ?? why only for junit jar now?
even more the test case might add some other jars in it.... you might want to log something into oracle or others, no?

PR #1611, opened 2018-04-06
TUP-19954 - Fixing spelling/grammatical mistakes

coheigea:


nrousseau:
note that normally we never merge translations manually, it's only done in babili... but still accept here.

PR #1625, opened 2018-04-10
TESB-21799: Fixed routes cannot use bean from reference project

yanbasic:
Fixed routes cannot use bean from reference project

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
@@ -33,6 +33,7 @@ import org.talend.core.model.general.Project;
 import org.talend.core.model.properties.Item;
 import org.talend.core.model.properties.JobletProcessItem;
 import org.talend.core.model.properties.ProcessItem;
+import org.talend.core.model.properties.ProjectReference;
 import org.talend.core.model.properties.Property;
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
```diff
@@ -904,12 +905,17 @@ public static boolean isRequiredBeans(IProcess process, Project project) {
...
-                    IProxyRepositoryService.class);
+            IProxyRepositoryService service = (IProxyRepositoryService) GlobalServiceRegister.getDefault()
+                    .getService(IProxyRepositoryService.class);
             ERepositoryObjectType beansType = ERepositoryObjectType.valueOf("BEANS"); //$NON-NLS-1$
             try {
                 IProxyRepositoryFactory factory = service.getProxyRepositoryFactory();
                 List<IRepositoryViewObject> all = factory.getAll(project, beansType);
+                List<ProjectReference> references = ProjectManager.getInstance().getCurrentProject()
+                        .getProjectReferenceList(true);
+                for (ProjectReference ref : references) {
```
zwxue: Should use recursive way since the structure of reference project system is hierarchical

zwxue: and should merge to master before backport to release branch

PR #1665, opened 2018-05-11
[TESB-20335] tESBConsumer not using http conduits when called from wi…

vbogatyrov:
…thin a subjob

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
@@ -19,9 +19,8 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 
-import org.eclipse.emf.common.util.EList;
+
 import org.talend.commons.exception.ExceptionHandler;
 import org.talend.commons.exception.PersistenceException;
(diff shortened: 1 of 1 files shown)


vbogatyrov:
Added several functions to ProcessUtils to check if given job is a 'ESB job'

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/ProcessUtils.java
```diff
@@ -975,4 +976,123 @@ public static boolean isRequiredPigUDFs(IProcess process, Project project) {
         return false;
     }
 
+    public static boolean isEsbJob(String processId, String version) {
```
nrousseau: ESB job can't have any joblet?
I'd suggest to try to put a flag somewhere during the generation inside the ProcessorUtilities if the child job have any of the ESB components, and use the list of components of generation instead of design.
Moverover here it's not optimized at all as it will reload the emf model again, while we already loaded the child before during the code generation. And this is not even checking if tRunjob is setup as independent or not.

vbogatyrov: Added flags ot ProcessUtils in order to not calculate whether job contains ESB components twice during code generation.  Also added joblet handling. I'm not sure what are "components of generation". Is it ProcessItem objects (in contrary to IRepositoryViewObject)?

PR #1680, opened 2018-05-21
fix(TUP-20124): NPE during changing TCOMP component repository value

cmeng-talend:
https://jira.talendforge.org/browse/TUP-20124


cmeng-talend:
Please also see https://github.com/Talend/tdi-studio-se/pull/2330 and https://github.com/Talend/tbd-studio-se/pull/789

PR #1726, opened 2018-06-12
fix(TDI-31645):Eliminate the invalid data type "DOUBLE PRECIS"

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-31645


pyzhou-talend:
I keep 
{code}				<dbType type="DOUBLE PRECIS">
					<talendType type="id_Double" default="true" />
					<talendType type="id_Float"/>
					<talendType type="id_BigDecimal"/>				
				</dbType>{code} on purpose in case regression.

PR #1729, opened 2018-06-12
fix(TUP-20248)Impact analysis doesn't work correctly against job in

jding-tlnd:
tRunJob component
https://jira.talendforge.org/browse/TUP-20248

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
@@ -301,18 +301,13 @@ public class RelationshipItemBuilder {
         }
         List<IRepositoryViewObject> allVersion = null;
         try {
-            allVersion = proxyRepositoryFactory.getAllVersion(itemId);
+            allVersion = proxyRepositoryFactory.getAllVersion(getAimProject(), itemId, true);
         } catch (PersistenceException e) {
             log.error(e);
         }
         Set<Relation> relations = new HashSet<Relation>();
         Set<Relation> currentItemKeySet = currentProjectItemsRelations.keySet();
         Set<Relation> refItemKeySet = referencesItemsRelations.keySet();
-        // in case there are some relations only with "Latest" version
-        Relation latestRelation = new Relation();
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
```diff
@@ -295,6 +295,58 @@ public void setProxyRepositoryFactory(IProxyRepositoryFactory proxyRepositoryFac
         return new ArrayList<Relation>(relations);
     }
 
+    public List<Relation> getAllVersionItemsRelatedTo(String itemId, String relationType, boolean withRefProject) {
+        if (!loaded) {
+            loadRelations();
+        }
+        List<IRepositoryViewObject> allVersion = null;
+        try {
+            allVersion = proxyRepositoryFactory.getAllVersion(itemId);
```
cmeng-talend: Better to use getProxyRepositoryFactory().getAllVersion(getAimProject(), itemId, true);
And please also help to change the usage at line 474;
The reason is that RelationshipItemBuilder can also be used for other projects instead of current project and reference projects during git merge.

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/relationship/RelationshipItemBuilder.java
```diff
@@ -295,6 +295,58 @@ public void setProxyRepositoryFactory(IProxyRepositoryFactory proxyRepositoryFac
...
+        Relation tmpRelation = new Relation();
+        tmpRelation.setId(itemId);
+        tmpRelation.setType(relationType);
+        for (IRepositoryViewObject object : allVersion) {
+            tmpRelation.setVersion(object.getVersion());
+            if (currentItemKeySet.contains(tmpRelation)) {
+                relations.addAll(currentProjectItemsRelations.get(tmpRelation));
+            }
+            // in case one side has relation but other side don't have
+            for (Relation base : currentItemKeySet) {
```
cmeng-talend: It looks like an exsiting function: RelationshipItemBuilder#getItemsHaveRelationWith(String itemId)

PR #1758, opened 2018-06-28
update cxf version

cmeng-talend:


nrousseau:
other solution was applied finally

PR #1771, opened 2018-07-05
bug fixed TDQ-12397 create IpropertyOrder interface

zshen-talend:


main/plugins/org.talend.model/src/main/java/org/talend/cwm/indicator/IPropertyOrder.java
```diff
@@ -0,0 +1,56 @@
+// ============================================================================
+//
+// Copyright (C) 2006-2016 Talend Inc. - www.talend.com
```
xqliu-talend: should be 2006-2018, need to make sure the java template is up to date

PR #1801, opened 2018-08-01
fix(TUP-20369): Mapper/Metadata tables for tMap in HTML document are missing in other than English.

jiezhang-tlnd:
https://jira.talendforge.org/browse/TUP-20369


i18n/plugins/org.talend.core.nl/src/main/java/messages_fr.properties
```diff
@@ -233,7 +233,7 @@ HTMLDocGenerator.context=Contexte :
 HTMLDocGenerator.context_list=Liste des contextes
 HTMLDocGenerator.creation=Cr�ation
 HTMLDocGenerator.description=Description
-HTMLDocGenerator.ended=termin�(e)
+HTMLDocGenerator.ended=termin�
```
wchen-talend: we should not change the .nl plugin

wchen-talend:
fix moved to https://github.com/Talend/tdi-studio-se/pull/2563

PR #1813, opened 2018-08-14
fix(TDI-40846): fix isDate condition

mbasiuk-talend:
Issue link - https://jira.talendforge.org/browse/TDI-40846
Date validation is incorrect in case of same date length but different value.

Remove length check, since equalsIgnoreCase condition is enough to verify compared date strings.

### DIFF
--- a/main/plugins/org.talend.librariesmanager/resources/java/routines/TalendDate.java
+++ b/main/plugins/org.talend.librariesmanager/resources/java/routines/TalendDate.java
@@ -249,12 +249,7 @@ public class TalendDate {
             return false;
         }
 
-        String formatDate = sdf.format(testDate);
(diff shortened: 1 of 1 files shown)


jzhao-talend:
@mbasiuk-talend  Did you check that whether your modification break old jobs?
https://jira.talendforge.org/browse/TDI-26262

mbasiuk-talend:
@jzhao-talend interesting jira... But the implementation of checking time zone is bad from my point of view. it made current bug to be real. Checking the length of input & formatted is not correct here. But it's the reason to think of improvement of it. Thanks for your finding!

jzhao-talend:
@mbasiuk-talend 
https://jira.talendforge.org/browse/TDI-24551

PR #1824, opened 2018-08-23
Jzhang/bugfix/tup 20511 need to limit the number of entry on the custom uri mapping json

jiezhang-tlnd:


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/CustomUriManager.java
```diff
@@ -92,13 +92,15 @@ private void saveResource(JSONObject customMap, String filePath, String fileName
...
 
-    public void saveCustomURIMap() {
+    public void saveCustomURIMap(boolean needWriteToJsonfile) {
         final RepositoryWorkUnit repositoryWorkUnit = new RepositoryWorkUnit(ProjectManager.getInstance().getCurrentProject(),
                 "Save custom maven uri map") {
 
             @Override
             public void run() throws PersistenceException, LoginException {
-                saveResource(customURIObject, getResourcePath(), CUSTOM_URI_MAP, false);
+            	if(needWriteToJsonfile) {
```
nrousseau: Why even execute a RWU if we don't save / commit anything ?

main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/CustomUriManager.java
```diff
@@ -159,7 +161,7 @@ public void reloadCustomMapping() {
         try {
             File file = new File(getResourcePath(), CUSTOM_URI_MAP);
             long modifyDate = file.lastModified();
-            if (modifyDate > lastModified) {
+            if (true) {
```
nrousseau: true ? little strange... it makes old check of code totally no use, better recheck with wchen for how this part was used before

jiezhang-tlnd: ok, I will recheck with wchen for this part that I 'm not familiar with.

PR #1864, opened 2018-09-26
bugfix(TUP-20617):Can't import the data models if the data containers

hwang-talend:
exist

### DIFF
--- a/main/plugins/org.talend.repository.items.importexport/src/main/java/org/talend/repository/items/importexport/handlers/imports/ImportBasicHandler.java
+++ b/main/plugins/org.talend.repository.items.importexport/src/main/java/org/talend/repository/items/importexport/handlers/imports/ImportBasicHandler.java
@@ -374,6 +374,7 @@ public class ImportBasicHandler extends AbstractImportExecutableHandler {
             }
 
             boolean isSameRepositoryType = true;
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.repository.items.importexport/src/main/java/org/talend/repository/items/importexport/handlers/imports/ImportBasicHandler.java
```diff
@@ -395,7 +396,9 @@ public boolean checkItem(ResourcesManager resManager, ImportItem importItem, boo
                         if (dbService != null && dbService.getExtraTypes().contains(repType)) {
                             repType = dbService.getExtraDBType(repType);
                         }
-                        isSameRepositoryType = isSameRepType(repType, importItem.getRepositoryType());
+                        ERepositoryObjectType importType = importItem.getRepositoryType();
+                        bothProcessType = allTypesOfProcess.contains(repType) && allTypesOfProcess.contains(importType);
+                        isSameRepositoryType = isSameRepType(repType, importType);
```
nrousseau: not really clean way to fix.. but only clean way seems full integration of mdm with DI/ESB repository.. which will take long time of course

PR #1870, opened 2018-09-29
fix(TUP-20758)Metadata connection to MSSQL with Integrated Security and

jding-tlnd:
TCP Dynamic ports
https://jira.talendforge.org/browse/TUP-20758


jding-tlnd:
Junit added

test/plugins/org.talend.core.runtime.test/src/org/talend/core/database/conn/DatabaseConnStrUtilTest.java
```diff
@@ -133,4 +133,24 @@ public void testAnalyseURLForVertica(){
         Assert.assertEquals(analyseURL.length, 6);
         Assert.assertEquals(analyseURL[4], "");
     }
+
+    @Test
+    public void testGetURLStringForMSSQL() {
+        String dbType = "Microsoft SQL Server";
+        String dbVersion = "MSSQL_PROP";
```
cmeng-talend: It is better to use EDatabaseVersion4Drivers and EDatabaseTypeName instead of const string

jding-tlnd: Solved, thanks

PR #1882, opened 2018-10-10
fix(TUP-20792):no table list after click select btn on amc.

hcyi:


main/plugins/org.talend.utils/src/main/java/org/talend/utils/sql/ConnectionUtils.java
```diff
@@ -516,4 +516,17 @@ public static boolean isNetezza(DatabaseMetaData databaseMetaData) {
         }
         return false;
     }
+
+    public static boolean isMysql(DatabaseMetaData metadata) throws SQLException {
+        boolean result = false;
+        try {
+            if (metadata != null && metadata.getDriverName() != null && metadata.getDatabaseProductName() != null
+                    && metadata.getDatabaseProductName().toUpperCase().indexOf("MYSQL") > -1) { //$NON-NLS-1$
```
wchen-talend: Can you remove the hard code ,to replace with product name in EDatabaseTypeName ?

hcyi: can not call class of  EDatabaseTypeName .

PR #1930, opened 2018-10-22
fix(TUP-21010):Jar name on classpath should be based on mvn uri

wchen-talend:


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/LocalLibraryManager.java
```diff
@@ -1258,6 +1258,14 @@ private void calculateModulesIndex(List<ModuleNeeded> modules, Map<String, Strin
                 } else {
                     libsToMavenUri.put(module.getModuleName(), mavenUrl);
                 }
+                // check if jar name is setup based on maven uri
+                String generatedName = MavenUrlHelper.generateModuleNameByMavenURI(mavenUrl);
+                if(!generatedName.equals(module.getModuleName())) {
+                    String context = module.getContext();
+                    String warning = "Module Name is expected as" + generatedName + ",but it is configured as "
```
nrousseau: Missing space to be readable here after the "as".
And i think we can add comments to avoid translation of this later, as it's more an internal string for dev.

PR #2023, opened 2018-11-16
fix(TUP-21077)Failed if import same job which contain tRuleSurvivorship

jiezhang-tlnd:
more than one time


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/repository/ERepositoryObjectType.java
```diff
@@ -1309,6 +1309,10 @@ public void setAllowPlainFolder(boolean isAllowPlainFolder) {
         this.isAllowPlainFolder = isAllowPlainFolder;
     }
 
+    public void setAllowMultiName(boolean isAllowMultiName) {
```
cmeng-talend: No need to add method here if we configure the plugin.xml dirrectly.

PR #2052, opened 2018-11-28
bugfix(TUP-21134):"Throw an error if the file already exist" is ticked

hwang-talend:
when import a 621 testcase for tFileoutputDelimited component


main/plugins/org.talend.core/src/main/java/org/talend/core/model/migration/AbstractBDJobMigrationTask.java
```diff
@@ -0,0 +1,37 @@
+// ============================================================================
```
zwxue: what's this class for? didn't see any class extends it

nrousseau: @zwxue 
I asked for this, to let BD team use this abstract class in the future. (and avoid to use a migration class that can impact as well DI jobs, as the migration class they use now seems for both DI/BD)

PR #2061, opened 2018-11-30
fix(TUP-20244)Guess query is not correct for tHiveInput if hivetable name is used a context value

jiezhang-tlnd:


main/plugins/org.talend.core/src/main/java/org/talend/core/model/metadata/query/AbstractQueryGenerator.java
```diff
@@ -417,7 +417,7 @@ protected String generateQueryDelegate() {
             final String tableNameWithDBAndSchema = getTableNameWithDBAndSchema(dbName, schemaName, tableName);
             String columnField = generateColumnFields(tableNameWithDBAndSchema);
             if (dbType == EDatabaseTypeName.HIVE) {
-                columnField = generateColumnFields(tableName);
+                columnField = generateColumnFields(SQL_START_CONNECTION + tableName + SQL_END_CONNECTION);
```
hcyi: only check it if context value

PR #2144, opened 2018-12-28
fix(TDI-41506):High CVE Vulnerability found in Talend Components

jzhao-talend:
* remove jars

For https://jira.talendforge.org/browse/TDI-41506


jzhao-talend:
The jar should be remove, but other team still not upgrade it. So keep it at present.

PR #2175, opened 2019-01-09
fix(TDQ-16280) Export 'commons-collections.3.2.2'

qiongli:


qiongli:
we can't update osgi dependency before upgrading Eclipse

PR #2189, opened 2019-01-11
fix(TUP-21205):Remove CryptoHelper from tcommon-studio-se +

ttao-talendbj:
tcommon-studio-ee

https://jira.talendforge.org/browse/TUP-21205


nrousseau:
CryptoHelper will be removed in another pull request from org.talend.utils, since this package is also used in TAC.

PR #2200, opened 2019-01-16
TUP-21308 Fix SSLUtils TLS configuration

kjwang-talend:
TUP-21308 Fix SSLUtils TLS configuration
https://jira.talendforge.org/browse/TUP-21308

### DIFF
--- a/main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
+++ b/main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
@@ -35,17 +35,6 @@ import javax.net.ssl.X509TrustManager;
  * DOC hcyi class global comment. Detailled comment
  */
 public class SSLUtils {
-
-    private static SSLUtils instance;
-
-    private String userWorkingDir = null;
-
-    private static HostnameVerifier hostnameVerifier;
-
-    private static KeyManager[] keystoreManagers;
-
-    private static TrustManager[] truststoreManagers;
-
     /**
      * {@value}
      * <p>
@@ -101,6 +90,16 @@ public class SSLUtils {
      * Accept all certification if don't setup tac.net.ssl.ClientTrustStore, the default value is <b>false</b>.
      */
     public static final String TAC_SSL_ACCEPT_ALL_IF_NO_TRUSTSTORE = "tac.net.ssl.AcceptAllIfNoTruststore"; //$NON-NLS-1$
+
+    private static SSLUtils instance;
+
+    private String userWorkingDir = null;
+
+    private static HostnameVerifier hostnameVerifier;
+
+    private static KeyManager[] keystoreManagers;
+
+    private static TrustManager[] truststoreManagers;
 
     /**
      * Get SSLUtils instance
@@ -203,10 +202,6 @@ public class SSLUtils {
 
     public HostnameVerifier getHostnameVerifier() {
         return hostnameVerifier;
-    }
-
-    public String getUserWorkingDir() {
-        return userWorkingDir;
     }
 
     // accept all certificate
--- a/main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
+++ b/main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
@@ -51,7 +51,7 @@ public class SSLUtils {
 
     public static final String TAC_SSL_ENABLE_HOST_NAME_VERIFICATION = "tac.net.ssl.EnableHostNameVerification"; //$NON-NLS-1$
 
-    public static final String TAC_SSL_FAIL_IF_NO_TRUSTSTORE = "tac.net.ssl.FailIfNoTruststore"; //$NON-NLS-1$
+    public static final String TAC_SSL_ACCEPT_ALL_IF_NO_TRUSTSTORE = "tac.net.ssl.AcceptAllIfNoTruststore"; //$NON-NLS-1$
 
     /**
      * 
@@ -67,7 +67,7 @@ public class SSLUtils {
         BufferedReader in = null;
         if (("https").equals(url.getProtocol())) {
             boolean openHttpHostNameVerification = Boolean
-                    .parseBoolean(System.getProperty(TAC_SSL_ENABLE_HOST_NAME_VERIFICATION));
+                    .parseBoolean(System.getProperty(TAC_SSL_ENABLE_HOST_NAME_VERIFICATION, Boolean.TRUE.toString()));
             final SSLSocketFactory socketFactory = getSSLContext(userDir).getSocketFactory();
             HttpsURLConnection httpsCon = (HttpsURLConnection) url.openConnection();
             httpsCon.setSSLSocketFactory(socketFactory);
@@ -95,7 +95,7 @@ public class SSLUtils {
             String trustStorePath = System.getProperty(TAC_SSL_CLIENT_TRUST_KEY);
             String keystorePass = System.getProperty(TAC_SSL_KEYSTORE_PASS);
             String truststorePass = System.getProperty(TAC_SSL_TRUSTSTORE_PASS);
-            boolean failIfNoTrustStore = Boolean.parseBoolean(System.getProperty(TAC_SSL_FAIL_IF_NO_TRUSTSTORE));
+            boolean acceptAllIfNoTrustStore = Boolean.parseBoolean(System.getProperty(TAC_SSL_ACCEPT_ALL_IF_NO_TRUSTSTORE));
             if (keystorePath == null) {
                 // if user does not set the keystore path in the .ini,we need to look for the keystore file under
                 // the root dir of product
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -51,7 +51,7 @@
 
     public static final String TAC_SSL_ENABLE_HOST_NAME_VERIFICATION = "tac.net.ssl.EnableHostNameVerification"; //$NON-NLS-1$
 
-    public static final String TAC_SSL_FAIL_IF_NO_TRUSTSTORE = "tac.net.ssl.FailIfNoTruststore"; //$NON-NLS-1$
+    public static final String TAC_SSL_ACCEPT_ALL_IF_NO_TRUSTSTORE = "tac.net.ssl.AcceptAllIfNoTruststore"; //$NON-NLS-1$
```
coheigea: It would be a good idea here to put in some Javadoc explaining what ths configuration option means (And that it is false by default)

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -67,7 +67,7 @@ public static String getContent(StringBuffer buffer, URL url, String userDir) th
         BufferedReader in = null;
         if (("https").equals(url.getProtocol())) {
             boolean openHttpHostNameVerification = Boolean
-                    .parseBoolean(System.getProperty(TAC_SSL_ENABLE_HOST_NAME_VERIFICATION));
+                    .parseBoolean(System.getProperty(TAC_SSL_ENABLE_HOST_NAME_VERIFICATION, Boolean.TRUE.toString()));
```
coheigea: I think a better variable description here would be to say "enableHostNameVerification" instead of "openHttpHostNameVerification"

kjwang-talend:
> Looks good - just two changes. Do all the tests pass with it?
Thanks coheigea. I will improve the code follow your suggestion. 
I have a question I think you can help me. You know,  There object "sslcontext" in class SSLUtils is single object. 
1. It is work well when  clientAuth="false" (That means client don't check the server's certification). 
2. But it will can't work when  clientAuth="true". I have checks it can connect to server sucessfuly only first time.  If I change the code, get a new sslcontext object everytime (That means I need to get a new SSLContext object and invoke "sslcontext.init(keystoreManagers, truststoreManagers, null) " again). It can work well.  Do you think it is acceptable? Or  I have some mistake on it, welcome give me some suggestion.
I am looking forward to your reply.

coheigea:
@kjwang-talend I don't understand why it would now be creating a new SSLContext object every time, whereas before it didn't. Are you sure that it didn't before? 

It sounds like the code that is causing SSLUtils needs to persist the SSLContext object rather than create it on every client call.

kjwang-talend:
> @kjwang-talend I don't understand why it would now be creating a new SSLContext object every time, whereas before it didn't. Are you sure that it didn't before?
> 
> It sounds like the code that is causing SSLUtils needs to persist the SSLContext object rather than create it on every client call.

Hi，coheigea
Thanks for your reply.
I have checked, use old code also can't work when clientAuth="true"(Only can work when clientAuth="false").  So I refactored all the code and modified the code follow your suggestion. Now It can work well for my test.
Please check it again. If you have any problem, please tell me.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -34,101 +35,130 @@
...
+     * {@value}
+     * <p>
+     * Open host name verification, the default value is <b>true</b>.
+     */
+    public static final String TAC_SSL_OPEN_HOST_NAME_VERIFICATION = "tac.net.ssl.OpenHttpHostNameVerification"; //$NON-NLS-1$
+
+    /**
+     * {@value}
+     * <p>
+     * Accept all certification if don't setup tac.net.ssl.ClientTrustStore, the default value is <b>false</b>.
```
coheigea: 'certificates' not 'certification'

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -34,101 +35,130 @@
...
+     */
+    public static final String TAC_SSL_ACCEPT_ALL_IF_NO_TRUSTSTORE = "tac.net.ssl.AcceptAllIfNoTruststore"; //$NON-NLS-1$
```
coheigea: The name of this variable is unclear. Maybe "AcceptAllCertsIfNoTruststore" insead.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -34,101 +35,130 @@
...
+     */
     public static final String TAC_SSL_TRUSTSTORE_PASS = "tac.net.ssl.TrustStorePass"; //$NON-NLS-1$
 
-    public static final String TAC_SSL_ENABLE_HOST_NAME_VERIFICATION = "tac.net.ssl.EnableHostNameVerification"; //$NON-NLS-1$
+    /**
+     * {@value}
+     * <p>
+     * Open host name verification, the default value is <b>true</b>.
+     */
+    public static final String TAC_SSL_OPEN_HOST_NAME_VERIFICATION = "tac.net.ssl.OpenHttpHostNameVerification"; //$NON-NLS-1$
```
coheigea: I think the old value of "tac.net.ssl.EnableHostNameVerification" is clearer.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -35,23 +41,173 @@
...
+            if (acceptAllCertsIfNoTrustStore) {
+                truststoreManagers = new TrustManager[] { new TrustAnyTrustManager() };
+            } else {
+                TrustManagerFactory tmf = TrustManagerFactory.getInstance("SunX509"); //$NON-NLS-1$
+                tmf.init((KeyStore) null);
+                truststoreManagers = tmf.getTrustManagers();
+            }
+        }
+
+        boolean openHttpHostNameVerification = Boolean
```
coheigea: Please change the variable name to "enableHostnameVerification"

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -35,23 +41,173 @@
...
+        }
+    }
+
+    public SSLContext getSSLContext() throws Exception {
+        SSLContext sslcontext = SSLContext.getInstance("SSL"); //$NON-NLS-1$
+        sslcontext.init(keystoreManagers, truststoreManagers, null);
+        return sslcontext;
+    }
+
+    public HostnameVerifier getHostnameVerifier() {
```
coheigea: Do we need this method? It's better not to expose internal information if we don't need to.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -35,23 +41,173 @@
...
+                truststoreManagers = new TrustManager[] { new TrustAnyTrustManager() };
+            } else {
+                TrustManagerFactory tmf = TrustManagerFactory.getInstance("SunX509"); //$NON-NLS-1$
+                tmf.init((KeyStore) null);
+                truststoreManagers = tmf.getTrustManagers();
+            }
+        }
+
+        boolean openHttpHostNameVerification = Boolean
+                .parseBoolean(System.getProperty(TAC_SSL_ENABLE_HOST_NAME_VERIFICATION, Boolean.TRUE.toString()));
```
coheigea: I would remove the cached hostnameVerifier and move this section down to the getContent method. The cached hostnameVerifier is not currently used anyway in getContent.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: ok. thanks. I found we still use this method on tamc-ee repositroy, So I restored old version. Anyway I improved it. We are using cached object now.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: Yes, we are still using it now. Please ref to the https://github.com/Talend/tcommon-studio-ee/pull/894/files -->ServletUtil.java and RestUtil.java

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: I improved it now. we are using cached hostnameVerifier now. Because the method  getHostnameVerifier(), I cached hostnameVerifier object. Why do you think "The cached hostnameVerifier is not currently used anyway in getContent." ?

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
coheigea: OK thanks!

PR #2207, opened 2019-01-17
fix(TUP-21505)be able to save jdbc connection even test fail

jiezhang-tlnd:


main/plugins/org.talend.core/src/main/java/org/talend/core/model/param/EConnectionParameterName.java
```diff
@@ -41,7 +41,11 @@
     GENERIC_PASSWORD("connection.userPassword.password"),//$NON-NLS-1$
     GENERIC_URL("connection.jdbcUrl"),//$NON-NLS-1$
     GENERIC_DRIVER_JAR("connection.driverTable"),//$NON-NLS-1$
-    GENERIC_DRIVER_CLASS("connection.driverClass");//$NON-NLS-1$
+    GENERIC_DRIVER_CLASS("connection.driverClass"), //$NON-NLS-1$
```
hcyi: No need add the new parameter

PR #2214, opened 2019-01-21
fix(TUP-17925)Sybase connection can check successfully even if a wrong db filled

jiezhang-tlnd:
db filled


hcyi:
Build running all the time?

PR #2218, opened 2019-01-23
Ttao/work item/tup 21143 generate random value for null password hiding code

ttao-talendbj:


main/plugins/org.talend.core/src/main/java/org/talend/core/model/utils/TalendTextUtils.java
```diff
@@ -495,27 +496,33 @@ public static boolean isEmptyString(final String str) {
...
-        
+
         if(!TalendQuoteUtils.isCommonString(password)){
             return password;
         }
 
-        int length = password.length() + 2;
+        SecureRandom random = new SecureRandom();
+        int passWordLength = random.nextInt();
+
```
coheigea: I think it's better if we simply print out a fixed number of "*" characters - say 8.

main/plugins/org.talend.core/src/main/java/org/talend/core/model/utils/TalendTextUtils.java
```diff
@@ -495,27 +496,33 @@ public static boolean isEmptyString(final String str) {
...
-        if (password == null) {
-            return PASS_COVER + PASS_COVER; // Means two quote
-        }
-
         if (ContextParameterUtils.containContextVariables(password)) {
             return password;
         }
-        
+
         if(!TalendQuoteUtils.isCommonString(password)){
```
coheigea: See if we can remove this if statement.

main/plugins/org.talend.core/src/main/java/org/talend/core/model/utils/TalendTextUtils.java
```diff
@@ -495,7 +494,6 @@ public static boolean isEmptyString(final String str) {
      * @return
      */
     public static String hidePassword(final String password) {
-
         if (ContextParameterUtils.containContextVariables(password)) {
             return password;
```
coheigea: I'm wondering if we need to call both "containContextVariables" and isCommonString here? Both methods end up calling quotePatternRegex.matcher in TalendQuoteUtils. Maybe "isCommonString" is enough?

PR #2239, opened 2019-02-12
fix(TUP-21880):Quick tour page is blank when launch studio using ESB

ttao-talendbj:
license

For the html file DocumentBuilderfactory, we shouldn't set
XMLConstants.FEATURE_SECURE_PROCESSING feature to true.
https://jira.talendforge.org/browse/TUP-21880


main/plugins/org.talend.presentation.onboarding/src/org/talend/presentation/onboarding/utils/OnBoardingUtils.java
```diff
@@ -256,7 +255,8 @@ public static String getCurrentSelectedPerspectiveId(IWorkbenchWindow wbWindow)
     public static Document convertStringToDocument(String htmlStr) {
         Document newDoc = null;
         try {
-            DocumentBuilderFactory factory = XmlUtils.getSecureDocumentBuilderFactory();
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
```
coheigea: It's always a good idea to setNamespaceAware(true) on DBF instances

PR #2252, opened 2019-02-21
TESB-24834 : fix NPE

sbouchet:


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
```diff
@@ -611,12 +616,15 @@ public void generateTemplates(boolean overwrite) throws Exception {
 
     protected ITalendProcessJavaProject getTalendJobJavaProject(JobInfo jobInfo) {
         IProcessor processor = jobInfo.getProcessor();
+        if (processor == null) {
+            processor = getJobProcessor();
+        }
```
yanbasic: Here is the real fixed NPE

PR #2279, opened 2019-03-07
TMDM-11900 Update schema of UpdateReport, add 'UUID' field for standalone mode.

sbliu:
https://jira.talendforge.org/browse/TMDM-11900

### DIFF
--- a/main/plugins/org.talend.repository.mdm/resources/UpdateReport.xsd
+++ b/main/plugins/org.talend.repository.mdm/resources/UpdateReport.xsd
@@ -11,6 +11,7 @@
 			<xsd:element maxOccurs="1" minOccurs="0" name="UserName" nillable="true" type="xsd:string"/>
 			<xsd:element maxOccurs="1" minOccurs="1" name="Source" nillable="false" type="xsd:string"/>
 			<xsd:element maxOccurs="1" minOccurs="1" name="TimeInMillis" nillable="false" type="xsd:string"/>
+			<xsd:element maxOccurs="1" minOccurs="1" name="UUID" nillable="false" type="xsd:string"/>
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.repository.mdm/resources/UpdateReport.xsd
```diff
@@ -11,6 +11,7 @@
 			<xsd:element maxOccurs="1" minOccurs="0" name="UserName" nillable="true" type="xsd:string"/>
 			<xsd:element maxOccurs="1" minOccurs="1" name="Source" nillable="false" type="xsd:string"/>
 			<xsd:element maxOccurs="1" minOccurs="1" name="TimeInMillis" nillable="false" type="xsd:string"/>
+			<xsd:element maxOccurs="1" minOccurs="1" name="UUID" nillable="false" type="xsd:string"/>
```
VictorLi-talendbj: Do we also need to have `UUID` in standalone?

sbliu: As task requirements, standalone mode only need add UUID element, cluster mode will change more, here only consider standalone mode change, so add UUID element .

sbliu: This PR could be merged at same time when related change in mdm part merged.

VictorLi-talendbj: Is it ok to have nillable="true" for UUID?

lpw82gf: No, it should be `false`,  we will set that value for sure. And if not, we can't switch from standalone to cluster.

PR #2292, opened 2019-03-13
TUP-21308 Fix SSLUtils TLS configuration

kjwang-talend:
TUP-21308 Fix SSLUtils TLS configuration
https://jira.talendforge.org/browse/TUP-21308


main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -89,83 +272,227 @@ public static String getContent(StringBuffer buffer, URL url, String userDir) th
...
+    }
+
+    private class AllowAllHostnameVerifier implements HostnameVerifier {
+
+        public boolean verify(String hostname, SSLSession session) {
+            return true;
+        }
+    }
+
+    private static class BrowserCompatibleHostnameVerifier implements HostnameVerifier {
```
coheigea: Why implement this here instead of re-using org.apache.http.conn.ssl.SSLSocketFactory.BROWSER_COMPATIBLE_HOSTNAME_VERIFIER as before?

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -15,43 +15,232 @@
...
+        }
+        if (truststorePass == null) {
+            // if user does not set the password in the talend.ini,we only can make it empty
+            // by
+            // default,but not sure the ssl can connect
+            truststorePass = ""; //$NON-NLS-1$
+        }
+
+        if (keystorePath != null) {
+            KeyManagerFactory kmf = KeyManagerFactory.getInstance("SunX509"); //$NON-NLS-1$
```
coheigea: Better to use KeyManagerFactory.getDefaultAlgorithm() here instead of SunX509

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -15,43 +15,232 @@
...
+            KeyStore ks = KeyStore.getInstance(KeyStore.getDefaultType());
+            ks.load(new FileInputStream(keystorePath), keystorePass == null ? null : keystorePass.toCharArray());
+            kmf.init(ks, keystorePass == null ? null : keystorePass.toCharArray());
+            keystoreManagers = kmf.getKeyManagers();
+        }
+
+        if (trustStorePath != null) {
+            TrustManagerFactory tmf = TrustManagerFactory.getInstance("SunX509"); //$NON-NLS-1$
```
coheigea: Better to use TrustManagerFactory.getDefaultAlgorithm() here instead of "SunX509"

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -15,43 +15,232 @@
...
+            tks.load(new FileInputStream(trustStorePath), truststorePass.toCharArray());
+            tmf.init(tks);
+            truststoreManagers = tmf.getTrustManagers();
+        }
+
+        if (truststoreManagers == null) {
+            if (acceptAllCertsIfNoTrustStore) {
+                truststoreManagers = new TrustManager[] { new TrustAnyTrustManager() };
+            } else {
+                TrustManagerFactory tmf = TrustManagerFactory.getInstance("SunX509"); //$NON-NLS-1$
```
coheigea: Same here.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -15,43 +15,232 @@
...
+                .parseBoolean(System.getProperty(TAC_SSL_ENABLE_HOST_NAME_VERIFICATION, Boolean.TRUE.toString()));
+        if (enableHostnameVerification) {
+            hostnameVerifier = new BrowserCompatibleHostnameVerifier();
+        } else {
+            hostnameVerifier = new AllowAllHostnameVerifier();
+        }
+    }
+
+    public SSLContext getSSLContext() throws Exception {
+        SSLContext sslcontext = SSLContext.getInstance("SSL"); //$NON-NLS-1$
```
coheigea: Use "TLS" instead of "SSL" here

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: The *org.talend.configurator* is supposed to be used to launch the studio from the licence and should not depend on anything. In order to keep the code are same with duplicated one, I implement it here.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: OK. Thanks

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: OK. Thanks

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: OK. Thanks

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
kjwang-talend: OK. Thanks

main/plugins/org.talend.utils/src/main/java/org/talend/utils/ssl/SSLUtils.java
```diff
@@ -15,43 +15,232 @@
...
+     */
+    public static final String TAC_SSL_ACCEPT_ALL_CERTS_IF_NO_TRUSTSTORE = "tac.net.ssl.AcceptAllCertsIfNoTruststore"; //$NON-NLS-1$
+
+    private HostnameVerifier hostnameVerifier;
+
+    private KeyManager[] keystoreManagers;
+
+    private TrustManager[] truststoreManagers;
+
+    private static Map<String, SSLUtils> userDirToInstanceMap = new HashMap<String, SSLUtils>();
```
coheigea: I'm wondering if the concept of a "userDir" map to SSLUtils is necessary? Before, the userDir was used as a location to read the keystores if not specified via properties. Then it (the SSLContext) was cached regardless of subsequent values of userDir. Perhaps we should just cache the SSLUtils instance instead as a static object?

kjwang-talend: Hi coheigea, Thanks for your reply. Because this class was used by other project. e.g AMC. I cached SSLUtils instance by key "userDir".  At least my change should not bring in new bugs. Because the compile order of different projects. I don't want to change the code in AMC here.  If you think I must to change it, please tell me.
Thanks again for your review.

PR #2296, opened 2019-03-15
fix(TPS-2975, TDI-41876) Importing a job from 6.2.1 to 7.1.1 is throwing error. Clean up user data 

dmytroshekera:
clean up user data for cases when default values wrapped with symbols :     ' " ( )
'000'
(000)
"NULL"

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
@@ -179,7 +179,7 @@ public final class MetadataToolAvroHelper {
                 defaultValue = getDefaultDoubleValue(defaultValue);
             } else if (JavaTypesManager.FLOAT.getId().equals(tt)) {
                 type = AvroUtils._float();
-                defaultValue = StringUtils.isEmpty((String)defaultValue) ? null : Float.parseFloat(defaultValue.toString());
+                defaultValue = getDefaultFloatValue(defaultValue);
             } else if (JavaTypesManager.BIGDECIMAL.getId().equals(tt)) {
                 // decimal(precision, scale) == column length and precision?
                 type = AvroUtils._decimal();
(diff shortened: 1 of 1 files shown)


gonchar-ivan:
https://jira.talendforge.org/browse/TDI-41876
added link to ticket

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -238,6 +238,93 @@ else if (JavaTypesManager.STRING.getId().equals(tt) || JavaTypesManager.FILE.get
...
+            return Float.parseFloat(result);
+        }
+    }
+
+    /*
+     * don't want to corrupt user data but should check couple cases
+     * in other case return the same expression - probably it is user's expression
+     */
+    private static String cleanupUserDefaultInputs(String inStr) {
+        if (inStr.length() > 2) {
```
gonchar-ivan: Do you think this check is necessary here?
I think about following input `"" ` (2 double quotes characters). It is not empty string and not null. This input will go to Lond.parseLong method and will cause 2nd exception.

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
```diff
@@ -238,6 +238,93 @@ else if (JavaTypesManager.STRING.getId().equals(tt) || JavaTypesManager.FILE.get
...
+            }
+            return Float.parseFloat(result);
+        }
+    }
+
+    /*
+     * don't want to corrupt user data but should check couple cases
+     * in other case return the same expression - probably it is user's expression
+     */
+    private static String cleanupUserDefaultInputs(String inStr) {
```
gonchar-ivan: I would suggest following method names:
trimQuotes
cleanString
cleanInput (maybe)

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/MetadataToolAvroHelper.java
dmytroshekera: empty string will cause NumberFormatException so it make no sense parse lower than 3 symbols. 
e.g. 
System.out.println(Long.parseLong(""));   // output    Exception in thread "main" java.lang.NumberFormatException: For input string: ""

dmytroshekera:
it was quick emergency case fix. 
General fix done. So no need in this PR

PR #2303, opened 2019-03-20
feat(TDI-41963): change jdbc driver version from 1.2.16.1027 to 1.2.2…

slushatel:
…0.1043


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/database/conn/version/EDatabaseVersion4Drivers.java
```diff
@@ -163,9 +163,9 @@
     MAPRDB(new DbVersion4Drivers(EDatabaseTypeName.MAPRDB, new String[] {})),
 
     REDSHIFT(new DbVersion4Drivers(EDatabaseTypeName.REDSHIFT, "redshift", "REDSHIFT", //$NON-NLS-1$ //$NON-NLS-2$
-            "redshiftJDBC42-no-awssdk-1.2.16.1027.jar")), //$NON-NLS-1$
+            "redshiftJDBC42-no-awssdk-1.2.20.1043.jar")), //$NON-NLS-1$
```
nrousseau: to be updated to redshift-jdbc42-no-awssdk-1.2.16.1027.jar (as there was a wrong case first on coponent)

cmeng-talend: updated

PR #2365, opened 2019-04-18
fix(TDQ-16899): force to update the Routine files.

msjian:


msjian:
in the past we had some performance issue because of this

msjian:
in the past we had some performance issue because of this

PR #2431, opened 2019-05-21
feat(TUP-23056):Avoid to manually deploy cibuilder / cloudpublisher for CI

zwxue:


main/plugins/org.talend.librariesmanager.ui/src/main/java/org/talend/librariesmanager/ui/startup/ShareMavenArtifactsOnStartup.java
```diff
@@ -104,4 +114,18 @@ public void shareToRepository(File file, MavenArtifact artifact) throws Exceptio
         // deployer.deploy(pomFile, artifact);
     }
 
+    private void addMojoArtifact(Map<ModuleNeeded, File> files, String groupId, String artifactId, String versionKey) {
+        String mvnUrl = MavenUrlHelper.generateMvnUrl(groupId, artifactId, VersionUtils.getMojoVersion(versionKey), null, null);
+        File file = null;
+        try {
+            file = TalendMavenResolver.getMavenResolver().resolve(mvnUrl);
```
wchen-talend: The resolve here will try to resolve it from remove .I think you need some maven url like 
mvn:localrepositories://@id=talend.local!org.talend.ci/builder-maven-plugin/<version> to 
revolve the path only in local maven.

wchen-talend: The resolve here will try to resolve it from remote repository .I think you need some maven url like
mvn:localrepositories://@id=talend.local!org.talend.ci/builder-maven-plugin/<version>
to revolve the path only in local maven.

PR #2451, opened 2019-05-28
fix(TUP-23196):HortonWorks dynamic distro cannot download commons codec

hcyi:


hcyi:
Waitting for esb team to check it.

PR #2472, opened 2019-06-05
fix(TDI-42356):change jtds jar to talend patched which include talend own class

jzhao-talend:
for patch:  https://jira.talendforge.org/browse/TPS-2855


jzhao-talend:
don't include wizard upgrade into patch

PR #2526, opened 2019-06-17
fix(TUP-23482)should avoid multi version slf4j-log4j jars in the job

jding-tlnd:
classpath
https://jira.talendforge.org/browse/TUP-23482

### DIFF
--- a/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/utils/MavenVersionHelper.java
+++ b/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/utils/MavenVersionHelper.java
@@ -13,8 +13,11 @@
 package org.talend.designer.maven.utils;
 
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
+import org.talend.core.runtime.maven.MavenConstants;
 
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/utils/MavenVersionHelper.java
```diff
@@ -25,4 +28,13 @@ public static int compareTo(String versionStr, String otherVersionStr) {
...
 
+    public static String getJarOriginalVersion(String jarName) {
+        String artifactId = jarName;
+        if (jarName.endsWith(MavenConstants.TYPE_JAR)) {
+            artifactId = jarName.substring(0, jarName.lastIndexOf(MavenConstants.TYPE_JAR) - 1);
+        }
+        String version = artifactId.substring(artifactId.lastIndexOf(SEGMENT) + 1);
+        return version;
+    }
+
```
wchen-talend: I don't think the get version function is good , it won't work for jar names like mysql-connector-java-5.1.30-bin.jar.
Your purpose is to remove duplicated version ,why don't use mvn url/gav ?

PR #2603, opened 2019-07-16
fix(TUP-23588)For tCreateTable, the Database Type is always Mysql after dragging and dropping from database Metadata

jiezhang-tlnd:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/designerproperties/RepositoryToComponentProperty.java
```diff
@@ -1697,6 +1697,11 @@ private static Object getDatabaseValue(DatabaseConnection connection, String val
                 return "STANDARD";
             }
         }
+        if (value.equals("DBTYPE")) {
```
hcyi: maybe you need check the code https://github.com/Talend/tcommon-studio-se/blob/520e4836739ecea5b28612ea683d480c7ef4fb31/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/metadata/designerproperties/RepositoryToComponentProperty.java#L903

PR #2641, opened 2019-08-02
TMDM-13805 Can read entity and navigator to previous/next page.

sbliu:
https://jira.talendforge.org/browse/TMDM-13805


main/plugins/org.talend.repository.mdm/src/main/java/org/talend/repository/mdm/ui/wizard/dnd/MDMSchema2TreeLinker.java
```diff
@@ -72,6 +72,8 @@
 
     private Comparator<LinkDescriptor<Item, Object, Tree, Object>> drawingLinksComparator;
 
+    private boolean setReadOnly = false;
+
```
hbhong: Using 'readOnly" instead of "setReadOnly" is better.

PR #2656, opened 2019-08-13
chore(TDQ-16113): code refactor

msjian:


main/plugins/org.talend.utils/src/main/java/org/talend/utils/string/StringUtilities.java
```diff
@@ -90,7 +92,14 @@ public static ReturnCode checkBalancedParenthesis(String input, char openingBloc
      * @return
      */
     public static String getRandomString(int length) {
-        return java.util.UUID.randomUUID().toString().replace("-", "").substring(8, 8 + length); //$NON-NLS-1$ //$NON-NLS-2$
+        String str = "abcdefghigklmnopkrstuvwxyzABCDEFGHIGKLMNOPQRSTUVWXYZ0123456789"; //$NON-NLS-1$
+        Random random = new SecureRandom();
+        StringBuilder sf = new StringBuilder();
+        for (int i = 0; i < length; i++) {
+            int number = random.nextInt(62);// 0~61
```
xqliu-talend: str.length() should be better than 62

PR #2692, opened 2019-08-29
fix : TDI-42776

rlecomte-talend:
Numeric sequence is now thread safe


clesaec:
Change approved.

I think you can also change :
```
public static void removeSequence(String seqName) {
        if (seq_Hash.containsKey(seqName)) {
            seq_Hash.remove(seqName);
        }
}
```

to
```
public static void removeSequence(String seqName) {
     seq_Hash.remove(seqName);
}
```
It has the same pb of synchronized, and remove operation should work even if key is absent.

rlecomte-talend:
removeSequence method changed as adviced

PR #2708, opened 2019-09-10
fix(TDI-42187): Fix Double Checked Locking in GlobalResource

wwang-talend:


main/plugins/org.talend.librariesmanager/resources/java/routines/system/GlobalResource.java
```diff
@@ -14,28 +14,31 @@
...
-                    if (tMultiTheadLockMap.get(key) == null) {
-                        tMultiTheadLockMap.put(key, new Object());
-                    }
-                }
-            }
-            return tMultiTheadLockMap.get(key);
-        }
-    }
+		public Object get(Object key) {
+			Object value = tMultiTheadLockMap.get(key);
```
coheigea: How about just something like this instead:

return tMultiThreadLockMap.computeIfAbsent(key, Object::new);

wwang-talend: thanks for your advice : @coheigea , just use the more readable method

PR #2709, opened 2019-09-11
fix(TPS-3358): [7.2.1] Job metrics: generate code too long + customization with system properties

wwang-talend:


wwang-talend:
after making sure with @nrousseau , this patch only works for talend inside for some test, so no need to backport it to patch/7.2.1 for integration patch to customer.

PR #2714, opened 2019-09-13
Rlecomte/fix/tdi 42188 harden deserialization

rlecomte-talend:
https://jira.talendforge.org/browse/TDI-42188


main/plugins/org.talend.librariesmanager/resources/java/routines/system/GetJarsToRegister.java
```diff
@@ -72,9 +72,9 @@ public String replaceJarPaths(String originalClassPathLine, String scheme, boole
 
         if (isNeedAddLibsPath(crcMapPath)) {
             java.util.Map<String, String> crcMap = null;
-            java.io.ObjectInputStream ois = new ObjectInputStream(new java.io.FileInputStream(crcMapPath));
-            crcMap = (java.util.Map<String, String>) ois.readObject();
-            ois.close();
+            try (java.io.ObjectInputStream ois = new ObjectInputStream(new java.io.FileInputStream(crcMapPath))) {
```
gonchar-ivan: I see your comment in Jira:
```
But the code uses a genuine ObjectInputStream, so check is enabled
```
What check do you mean here?
I didn't get why class name validation is not required here and it is ok to use ObjectInputStream here.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/NoHeaderObjectInputStream.java
```diff
@@ -25,10 +28,25 @@
...
+    /**
+     * If {@link #expectedTypes} is set, flags if the class has already been checked
+     */
+    private boolean valid = false;
+
     public NoHeaderObjectInputStream(InputStream in) throws IOException {
         super(in);
     }
 
+    public NoHeaderObjectInputStream(InputStream in, Class<?>... expectedTypes) throws IOException {
```
gonchar-ivan: I see minor issue here.
NoHeaderObjectInputStream instance state can be changed after instance was initialized.
e.g:
```
Class<?>[] whitelist = new Class[] {String.class, Integer.class};
NoHeaderObjectInputStream stream = new NoHeaderObjectInputStream(in, whitelist);
whitelist[0] = SomeOther.class;
```
Not sure it is necessary to fix.
Looks like here is a good example
https://github.com/apache/commons-io/blob/master/src/main/java/org/apache/commons/io/serialization/ValidatingObjectInputStream.java#L111-L115
(We don't need FullClassNameMatcher, but the idea is to make a copy of the input array)

main/plugins/org.talend.librariesmanager/resources/java/routines/system/NoHeaderObjectInputStream.java
```diff
@@ -44,4 +62,18 @@ protected NoHeaderObjectInputStream() throws IOException, SecurityException {
     protected void readStreamHeader() throws IOException, StreamCorruptedException {
         // don't need to check the header
     }
+
+    @Override
+    protected Class<?> resolveClass(ObjectStreamClass desc) throws IOException, ClassNotFoundException {
+        if (expectedTypes != null && !valid) {
```
gonchar-ivan: As I see we do a validation only for the first class (and I have a question why we don't do a validation each time)
Do I understand it correctly?
The first class passed to validation will be TraceDataBean or TraceStatusBean.
Next classes passed to this method will be classes of fields.
We don't want to list classes of all fields, so we check only top level class (the first).
Is it correct?

main/plugins/org.talend.librariesmanager/resources/java/routines/system/GetJarsToRegister.java
rlecomte-talend: The check I'm talking about is the header check (empty overidded method in NoHeaderObjectInputStream). In GetJarsToRegister, the deserialization is made by a ObjectInputStream, so stream header is read ; version and magic number of the class is checked.
The class name validation is not easy to do in the GetJarsToRegister because all we know about the serialized is that it ias a Map<String, String>. But it can be a HashMap,a TreeMap, a Hashtable or any other custom implementation of the Map, so what class names should we accept ? In Runtrace, it was possible because the object could only be of 2 types (only 2 implementations of the TraceBean interface).

rlecomte-talend: But doing that, you resolve the class (Class.forName) to check if it's safe to resolve the class.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/NoHeaderObjectInputStream.java
rlecomte-talend: That's right. When I was developping this fix, I encountered this issue, that's the way I found to solve it.
But that means that an instance of NoHeaderObjectInputStream can only be used once. Otherwize, the next readObject won't check anything.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/NoHeaderObjectInputStream.java
```diff
@@ -25,10 +28,25 @@
...
+     */
+    private boolean valid = false;
+
     public NoHeaderObjectInputStream(InputStream in) throws IOException {
         super(in);
     }
 
+    public NoHeaderObjectInputStream(InputStream in, Class<?>... expectedTypes) throws IOException {
+        this(in);
+        this.expectedTypes = expectedTypes;
```
coheigea: Like @gonchar-ivan said above, the expectedTypes parameter should be defensively copied to make sure that the array can't be modified externally to the class.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/NoHeaderObjectInputStream.java
```diff
@@ -44,4 +62,18 @@ protected NoHeaderObjectInputStream() throws IOException, SecurityException {
     protected void readStreamHeader() throws IOException, StreamCorruptedException {
         // don't need to check the header
     }
+
+    @Override
+    protected Class<?> resolveClass(ObjectStreamClass desc) throws IOException, ClassNotFoundException {
+        if (expectedTypes != null && !valid) {
+            if (Arrays.stream(expectedTypes)
```
coheigea: Easier to write: Arrays.stream(expectedTypes).anyMatch(c -> c.getName().equals(desc.getName()); ?

main/plugins/org.talend.librariesmanager/resources/java/routines/system/GetJarsToRegister.java
coheigea: I think this might be the only class that writes the crcMap, and it uses a HashMap:

https://github.com/Talend/jobserver/blob/32818cae5f79759d079945b537575c7e4a1bbb93/org.talend.remote.jobserver.commons/src/main/java/org/talend/remote/jobserver/commons/utils/io/Unzipper.java

rlecomte-talend: We could eventually test that the Map implementation is a HashMap but :
- that would break the use of an interface, meaning that if someone changes the Map implemetation, we'll have to change or complete the test
- We must then verify the type of each object, verifying that each key and value of the map is a String (possibility to use a HashMap to send some malicious Object as a value, and the  NoHeaderObjectInputStream can't check it now).
Is it really worth it ?

main/plugins/org.talend.librariesmanager/resources/java/routines/system/NoHeaderObjectInputStream.java
rlecomte-talend: Done with :
```
public NoHeaderObjectInputStream(InputStream in, Class<?>... expectedTypes) throws IOException {
        this(in);
        if (expectedTypes != null) {
            this.expectedTypes = new Class<?>[expectedTypes.length];
            System.arraycopy(expectedTypes, 0, this.expectedTypes, 0, expectedTypes.length) ;
        }
    }
```

main/plugins/org.talend.librariesmanager/resources/java/routines/system/GetJarsToRegister.java
rlecomte-talend: Furthemore, this code also appears in some javajet file : 
https://github.com/search?q=org%3ATalend+crcMap%22&type=Code

fdesiles: Seen with @pteyssier, modification will apply only to this scope. @rlecomte-talend made the changes and re-requested the review from @coheigea.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/GetJarsToRegister.java
```diff
@@ -72,9 +73,9 @@ public String replaceJarPaths(String originalClassPathLine, String scheme, boole
 
         if (isNeedAddLibsPath(crcMapPath)) {
             java.util.Map<String, String> crcMap = null;
-            java.io.ObjectInputStream ois = new ObjectInputStream(new java.io.FileInputStream(crcMapPath));
-            crcMap = (java.util.Map<String, String>) ois.readObject();
-            ois.close();
+            try (java.io.ObjectInputStream ois = new NoHeaderObjectInputStream(new java.io.FileInputStream(crcMapPath), HashMap.class)) {
```
coheigea: Thinking about this some more, I think the change to GetJarsToRegister is not required. It's reading a file from the local filesystem, which is only written out by Talend code, so there is not much risk involved. So here I think it's fine just to leave in the try-with-resources change. The RunTrace is fine though.

coheigea:
This PR should be closed

gonchar-ivan:
A copy of this PR was created and merged https://github.com/Talend/tcommon-studio-se/pull/2884
This PR will be closed.
Thanks Colm!

PR #2754, opened 2019-10-10
fix(TDI-43035): use only main connection category

mbasiuk-talend:
https://jira.talendforge.org/browse/TDI-43035

Use only MAIN Connection Category and avoid returning Lookup.

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
@@ -684,16 +684,15 @@ public class NodeUtil {
             }
             List<? extends IConnection> listInConns = node.getIncomingConnections();
             if (listInConns != null && listInConns.size() > 0) {
-                String retResult = getPrivateConnClassName(listInConns.get(0));
-                if (retResult == null) {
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
```diff
@@ -684,16 +684,15 @@ public static String getPrivateConnClassName(final IConnection conn) {
             }
             List<? extends IConnection> listInConns = node.getIncomingConnections();
             if (listInConns != null && listInConns.size() > 0) {
-                String retResult = getPrivateConnClassName(listInConns.get(0));
-                if (retResult == null) {
-                    return conn.getName();
-                } else {
-                    return retResult;
+                for (IConnection connection : listInConns) {
+                    if (EConnectionCategory.MAIN == connection.getLineStyle().getCategory()) {
```
wwang-talend: getCategory() method is @Deprecated method, and i don't know what EConnectionCategory.MAIN/OTHER mean, and also advice to use this logic : 
```
if(connection.getLineStyle()!=EConnectionType.FLOW_REF) {
```
I think it is more safe and also can find forward even some other connection like TABLE / FLOW_MERGE Links

PR #2760, opened 2019-10-11
fix(TBD-9333): add new methods to get previous node

AlixMetivier:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/INode.java
```diff
@@ -182,6 +182,10 @@
     public List<? extends IConnection> getIncomingConnections(EConnectionType connectionType);
 
     public List<? extends IConnection> getOutgoingConnections(String connectorName);
+
```
lbourgeois: I would add method Javadoc with _Get previous component from first incoming connection_

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/INode.java
```diff
@@ -182,6 +182,10 @@
     public List<? extends IConnection> getIncomingConnections(EConnectionType connectionType);
 
     public List<? extends IConnection> getOutgoingConnections(String connectorName);
+    
+    public IComponent getPreviousComponent();
+    
+    public IComponent getPreviousComponent(int index);
```
lbourgeois: I would add method Javadoc with _Get previous component by incoming connection index_ + parameter desc

PR #2781, opened 2019-10-23
Hengzhao/feat/tup 21079 allow to setup a different nexus for proxy of official nexus

hzhao-talendbj:


main/plugins/org.talend.librariesmanager.ui/LicenseText.json
```diff
@@ -0,0 +1,386 @@
+[{
```
wchen-talend: Can you please create a single file for each license text ? 
it is difficult to maintain a json file if any new license added/text change.

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/nexus/TalendLibsServerManager.java
```diff
@@ -61,7 +61,15 @@
...
     public static final String KEY_SOFTWARE_UPDATE_REPOSITORY = "repositoryID";//$NON-NLS-1$
 
-    public static final String TALEND_LIB_SERVER = "https://talend-update.talend.com/nexus/";//$NON-NLS-1$
+    public static String TALEND_LIB_SERVER = "https://talend-update.talend.com/nexus/";//$NON-NLS-1$
+
+    public static final String NEXUS_PROXY_URL = "nexus.proxy.url";
+
+    public static final String NEXUS_PROXY_TYPE = "nexus.proxy.type";
+
+    public static final String NEXUS_PROXY_USERNAME = "nexus.proxy.username";
```
wchen-talend: user/password is never used ?

main/plugins/org.talend.librariesmanager.ui/src/main/java/org/talend/librariesmanager/utils/ChangeHtmlToJavaUtil.java
```diff
@@ -0,0 +1,151 @@
+package org.talend.librariesmanager.utils;
```
wchen-talend: I think you make it too complex , just simply use file . and you can display the file content in the brower .

hzhao-talendbj: According to @cmeng-talend,  some OS can't use Browser ,so use Text  instead,and read the license text to a String and set it into the Text to show in dialog

PR #2806, opened 2019-11-01
feat(TUP-24862) encryption key rotation

bhe-talendbj:
feat(TUP-24862) encryption key rotation

https://jira.talendforge.org/browse/TUP-24862


main/plugins/org.talend.librariesmanager/resources/java/routines/system/PasswordEncryptUtil.java
```diff
@@ -38,17 +38,24 @@ public static String encryptPassword(String input) throws Exception {
...
+        return sb.toString();
     }
 
     public static String decryptPassword(String input) {
         if (input == null || input.length() == 0) {
             return input;
         }
-        if (input.startsWith(PREFIX_PASSWORD) && input.endsWith(POSTFIX_PASSWORD)) {
+        if (input.startsWith(PREFIX_PASSWORD + KEY_FIXED + ":")) {
+            int idx = input.lastIndexOf(':');
```
coheigea: I think it would be better here to use something like:
int idx =  (PREFIX_PASSWORD + KEY_FIXED + ":").getLength()

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioKeySource.java
```diff
@@ -0,0 +1,156 @@
...
+            LOGGER.error(e);
+        }
+
+        // load from file set in system property, so as to override default keys
+        String keyPath = System.getProperty(StudioKeysFileCheck.ENCRYPTION_KEY_FILE_SYS_PROP);
+        if (keyPath != null) {
+            File keyFile = new File(keyPath);
+            if (keyFile.exists()) {
+                try (InputStream fi = new FileInputStream(keyFile)) {
+                    allKeys.load(fi);
```
coheigea: How come we load all keys here, but only load keys starting with KEY_SYSTEM_PREFIX from the system properties below?

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioKeySource.java
```diff
@@ -0,0 +1,156 @@
...
+public class StudioKeySource implements KeySource {
+
+    private static final Logger LOGGER = Logger.getLogger(StudioKeySource.class);
+
+    public static final String KEY_SYSTEM_PREFIX = "system.encryption.key.v";
+
+    public static final String KEY_FIXED = "routine.encryption.key";
+
+    // TODO: this fixed key will be removed shortly
+    private static final String FIXED_ENCRYPTION_KEY_DATA = "Talend_TalendKey";
```
coheigea: Is there a JIRA to track removing this key?

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioKeySource.java
```diff
@@ -0,0 +1,156 @@
...
+
+    private String systemKeyForEncrypt;
+
+    private StudioKeySource(Properties allKeys, String keyName, boolean isMaxVersion) {
+        this.availableKeys = allKeys;
+        this.keyName = keyName;
+        this.isEncrypt = isMaxVersion;
+
+        // get highest version of system encryption key, this key will be used to encrypt data
+        this.systemKeyForEncrypt = availableKeys.stringPropertyNames().stream().filter(e -> e.startsWith(KEY_SYSTEM_PREFIX))
```
coheigea: Same point I made here on the other PR - there's no need to have a separate property for encryption, just use keyName

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioKeySource.java
```diff
@@ -0,0 +1,156 @@
...
+            LOGGER.warn("Parse version of encryption key error, key: " + keyName);
+        }
+        return 0;
+    }
+
+    /**
+     * Get key name corresponding to the key source
+     */
+    public String getKeyName() {
+        if (this.isEncrypt && this.keyName.startsWith(KEY_SYSTEM_PREFIX)) {
```
coheigea: Move this if statement into the constructor and just return keyName here.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioEncryption.java
```diff
@@ -13,53 +13,57 @@
...
 
-    private static final String POSTFIX_PASSWORD = "]"; //$NON-NLS-1$
+    private static final Pattern REG_ENCRYPTED_DATA_ROUTINE = Pattern.compile("^enc\\:routine\\.encryption\\.key\\:\\p{Print}+");
 
-    // Encryption key property names
-    private static final String KEY_SYSTEM = "system.encryption.key.v1";
+    private static final Pattern REG_ENCRYPTED_DATA_M3 = Pattern.compile("^ENC\\:\\[\\p{Print}+\\]$");
+
+    // Encryption key name shipped in M3
+    private static final String KEY_SYSTEM_M3 = StudioKeySource.KEY_SYSTEM_PREFIX + "1";
```
coheigea: Do we really need to support M3? It's not very efficient having all these reg-exs.

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioKeySource.java
bhe-talendbj: jira:TUP-22966

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioKeySource.java
bhe-talendbj: Done

main/plugins/org.talend.utils/src/main/java/org/talend/utils/security/StudioEncryption.java
bhe-talendbj: Yes, discussed it with  @nrousseau , we need to support jobs created in M3 studio.

PR #2811, opened 2019-11-06
feat(TUP-24717):Activate/Deactivate plugins when install patch/update

zwxue:


main/plugins/org.talend.updates.runtime/src/main/java/org/talend/updates/runtime/utils/UpdateTools.java
```diff
@@ -0,0 +1,279 @@
...
+            throws Exception {
+        if (installingPatchFolder != null && installingPatchFolder.exists()) {
+            File carFolder = new File(installingPatchFolder, ITaCoKitUpdateService.FOLDER_CAR);
+            TaCoKitCarUtils.installCars(carFolder, monitor, cancellable);
+        }
+        return true;
+    }
+
+    public static void syncLibraries(File installingPatchFolder) throws IOException {
+        // sync to product lib/java
```
wchen-talend: lib/java should not be used anymore , I don't think we will need any lib in patch and synchronize to product.

zwxue: it's a part of tcompv0 patch installation function, it's still working. just keep this in case.

PR #2815, opened 2019-11-06
feat(TUP-24849)Remove jars from the studio

jding-tlnd:
https://jira.talendforge.org/browse/TUP-24849


main/plugins/org.talend.libraries.apache.cassandra/plugin.xml
```diff
@@ -1,75 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
wchen-talend: There is librariesNeeded extension in this plugin.xml ,need to add back , otherwise , component will not able to find the jars.

jding-tlnd: Solved, thanks

cmeng-talend:
Question:
Do we need to keep jars configured in the pom.xml?
https://github.com/Talend/tcommon-studio-se/blob/jding/feat/TUP-24849_Remove_jars_from_the_studio/main/plugins/org.talend.libraries.apache.cassandra/pom.xml#L23-L60
These jars will be added to the plugin during building, if no need them, it's better to remove them too, then remove the lib folder from build.properties
https://github.com/Talend/tcommon-studio-se/blob/jding/feat/TUP-24849_Remove_jars_from_the_studio/main/plugins/org.talend.libraries.apache.cassandra/build.properties#L4

PR #2836, opened 2019-11-13
TUP-25128 Studio miss item properties when change item version(for routine/json items)

kjwang-talend:
TUP-25128 Studio miss item properties when change item version(for routine/json items)
https://jira.talendforge.org/browse/TUP-25128


kjwang-talend:
This pull request to resolve issue : https://jira.talendforge.org/browse/TUP-25128

PR #2845, opened 2019-11-18
bugfix(TUP-24953):Nexus 3 gets populated by crypto-utils.jar

hwang-talend:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/routines/RoutineLibraryMananger.java
```diff
@@ -98,7 +98,7 @@ public void initializeSystemLibs() {
                                     if(fileUrl != null){
                                     	if (!"file".equals(fileUrl.getProtocol())) throw new IllegalArgumentException();
                                     	URI fileUri = new File(fileUrl.getFile()).toURI();
-                                        libManagerService.deploy(fileUri);
+                                        libManagerService.deploy(fileUri, null, false);
```
wchen-talend: For system routine , the modification is ok , but if customer routine there might be problem if not share to nexus.
Maybe we need some check like 
For release jar : LocalLibraryManager.resolveJar(...), if already installed,  don't deploy again.
For snapshot jars : LocalLibraryManager.resolveJar(...) , if jar already installed compare the sha1 of the two jars LocalLibraryManager.isSameFile(...) ,  deploy again only if the two jars are not the same one.

PR #2858, opened 2019-11-25
fix(TDI-43211): not find the slf4j-api-1.7.25.jar in lib dir

wwang-talend:


wwang-talend:
don't merge please

wwang-talend:
seems nexcus is down

PR #2862, opened 2019-11-26
Jzhang/feat/tup 24651 add choice between log4j and log4j2 in project settings

jiezhang-tlnd:


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/Log4jUtil.java
```diff
@@ -1,10 +1,8 @@
 package org.talend.core.repository.utils;
 
-import org.talend.core.PluginChecker;
-
 public class Log4jUtil {
 
     public static boolean isEnable() {
-        return PluginChecker.isTIS();
+        return true;
```
wchen-talend: This function is called when create log4j setting page,
I think there will be problem and  it will create the enable part for TOS

PR #2882, opened 2019-12-04
feat(TDI-42591): bigquery dynamic column support

chmyga:
* Update bigquery mapping xml


main/plugins/org.talend.core.runtime/mappings/mapping_BigQuery.xml
```diff
@@ -60,13 +71,16 @@
                     <dbType type="TIME" />
                     <dbType type="TIMESTAMP" />
                     <dbType type="ARRAY" />
-                    <dbType type="RECORD" />
+                    <dbType type="GEOGRAPHY" />
```
dmytro-sylaiev: you still have Record to String mapping on <dbToTalendTypes>
And no mention of RECORD type in talendToDbTypes
Can you please elaborate on this?

PR #2893, opened 2019-12-06
fix(TPS-3611):back port some code on patch/7.0.1 from TUP-21077 to fix edit rules can not export again

zshen-talend:


zshen-talend:
buildme

PR #2913, opened 2019-12-12
feat(TUP-24523): Online Help in studio - refactor help features

kjwang-talend:
https://jira.talendforge.org/browse/TUP-24523


main/features/org.talend.help.feature/feature.xml
```diff
@@ -10,7 +9,8 @@
   </url>  
   <requires>
     <import feature="org.eclipse.help" version="0.0.0" match="greaterOrEqual"/>
+    <import feature="org.talend.help.di.feature" version="0.0.0" match="greaterOrEqual"/>
+    <import feature="org.talend.help.bd.feature" version="0.0.0" match="greaterOrEqual"/>
+    <import feature="org.talend.help.esb.feature" version="0.0.0" match="greaterOrEqual"/>
```
nrousseau: better don't have any import of org.talend.help here.
We want the help of component to be totally separated.
If add here, it will always add in the build still. (same as before)

PR #2934, opened 2019-12-20
TESB-27139 Backport from master branch

dsergent:
"Could not find the context Default" is shown in runtime when deploy a
route into runtime


csuntalendbj:
Hi, Please check 
https://github.com/Talend/tdi-studio-se/blob/2ad1c00b636673c2c10251c8198a97875332d6b6/main/plugins/org.talend.designer.codegen/jet_stub/header.javajet#L142
we are still using isEsbJob method, so please revert that method

PR #2940, opened 2019-12-24
feat(TUP-24722):Force each studio on the same project to install the same patch

zwxue:


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/utils/MavenHelper.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static String getArtifactPath(MavenArtifact artifact) throws CoreException {
+        ArtifactRepository localRepository = MavenPlugin.getMaven().getLocalRepository();
+        String repositoryPath = localRepository.getBasedir();
+        String artifactPath = MavenPlugin.getMaven().getArtifactPath(localRepository, artifact.getGroupId(),
+                artifact.getArtifactId(),
+                artifact.getVersion(), artifact.getType(), artifact.getClassifier());
+        return repositoryPath + "/" + artifactPath;
+    }
+
```
wchen-talend: It seems it is the same as existing function 
PomUtil.getAbsArtifactPathAsCP(MavenArtifact artifact)/getAbsArtifactPathAsCP(MavenArtifact artifact)

zwxue: resolved

PR #2985, opened 2020-01-08
TESB-27819:Publishing artifacts to cloud fail with Studio M5 / Java 11

csuntalendbj:


csuntalendbj:
AFAIK, JDK 8 do have related classes, but JDK 11 the related classes were removed by the official, we need to manually add it into the classpath. Also some of the jiras which we handled  with JDK 11 were all manually add that dependencies.

dsergent:
> AFAIK, JDK 8 do have related classes, but JDK 11 the related classes were removed by the official, we need to manually add it into the classpath. Also some of the jiras which we handled with JDK 11 were all manually add that dependencies.

This was not really the question, it s OK that the bundle needs theses libs. Question is, why we commit jar files in GitHub? 'lib' folder should be populated with maven build and never commited (added to .gitignore), WDYT?

csuntalendbj:
> > AFAIK, JDK 8 do have related classes, but JDK 11 the related classes were removed by the official, we need to manually add it into the classpath. Also some of the jiras which we handled with JDK 11 were all manually add that dependencies.
> 
> This was not really the question, it is OK that the bundle needs these libs. Question is: why do we commit jar files in GitHub? 'lib' folder should be populated with maven build and never commited (added to .gitignore), WDYT?

Yes, for our ESB plugins, we did that behavior, but it is not an ESB's plugin, and it is not maintenance by ESB team, I think we could suggest them to do that. WDYT?

PR #2999, opened 2020-01-15
bugfix(TUP-25827) Upgrade commons-codec, commons-io and ant dependencies

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-25827

### DIFF
--- a/main/plugins/org.talend.utils/pom.xml
+++ b/main/plugins/org.talend.utils/pom.xml
@@ -26,6 +26,11 @@
 			<version>2.6</version>
 		</dependency>
 		<dependency>
+			<groupId>org.apache.ant</groupId>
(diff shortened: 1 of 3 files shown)


xldai:
Maybe better upgrade the commons-io to the latest 2.6, as the jobserver already had https://github.com/Talend/jobserver/blob/master/pom.xml#L36, also the commons-io/2.6 has been used all over the ESB Runtime already.

bhe-talendbj:
> Maybe better upgrade the commons-io to the latest 2.6, as the jobserver already had https://github.com/Talend/jobserver/blob/master/pom.xml#L36, also the commons-io/2.6 has been used all over the ESB Runtime already.

Done. thanks.

PR #3018, opened 2020-01-22
fix(TBD-9863): jar conflict on CDH 6.x

aparent-talend:
hive-exec jar is loaded before commons-lang3 but hive-exec is a fatjar that contains an old version of commons-lang3
hence a lot of job are failing because this breaks all the tFileOutput components

https://jira.talendforge.org/browse/TBD-9863


aparent-talend:
Waiting for approvals from TUP team.

PR #3025, opened 2020-01-28
fix(TDI-43295): Support Oracle 19c in Studio DI

sbovsunovskyi:
https://jira.talendforge.org/browse/TDI-43295


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/database/conn/version/EDatabaseVersion4Drivers.java
```diff
@@ -30,6 +30,8 @@
     ACCESS_2003(new DbVersion4Drivers(EDatabaseTypeName.ACCESS, "Access 2003", "Access_2003")), //$NON-NLS-1$ //$NON-NLS-2$
     ACCESS_2007(new DbVersion4Drivers(EDatabaseTypeName.ACCESS, "Access 2007", "Access_2007")), //$NON-NLS-1$ //$NON-NLS-2$
     // oracle
+    ORACLE_19(new DbVersion4Drivers(new EDatabaseTypeName[] { EDatabaseTypeName.ORACLEFORSID, EDatabaseTypeName.ORACLESN,
+            EDatabaseTypeName.ORACLE_OCI, EDatabaseTypeName.ORACLE_CUSTOM }, "Oracle 19", "ORACLE_19", "ojdbc8-19.3.0.0.jar")),
```
gonchar-ivan: I saw there are several versions of JDBC driver on maven central.
Do you know a difference between them?
1. g https://mvnrepository.com/artifact/com.oracle.ojdbc/ojdbc8_g/19.3.0.0
2. dms https://mvnrepository.com/artifact/com.oracle.ojdbc/ojdbc8dms/19.3.0.0
3. dms_g https://mvnrepository.com/artifact/com.oracle.ojdbc/ojdbc8dms_g/19.3.0.0
4. without suffix https://mvnrepository.com/artifact/com.oracle.ojdbc/ojdbc8/19.3.0.0

sbovsunovskyi: I see that we always use drivers without any letter in it's name.

gonchar-ivan: Stas clarified meaning of different versions:
dms - is dynamic monitoring
g - is javac option which generates different debug information in compiled classes
This is not needed. So, version without suffix should be used

PR #3033, opened 2020-02-03
bugfix(TUP-24536) Studio is vulnerable to "Zip Slip" style attacks

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-24536

### DIFF
--- a/main/plugins/org.talend.utils/src/main/java/org/talend/utils/io/FilesUtils.java
+++ b/main/plugins/org.talend.utils/src/main/java/org/talend/utils/io/FilesUtils.java
@@ -794,10 +794,12 @@ public final class FilesUtils {
      * @param destPathStr destination file path
      */
     public static void validateDestPath(String targetDir, String destPathStr) throws IOException {
-        Path destPath = Paths.get(destPathStr);
-        Path destPathNormalized = destPath.normalize();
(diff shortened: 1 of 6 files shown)


main/plugins/org.talend.utils/src/main/java/org/talend/utils/io/FilesUtils.java
```diff
@@ -785,6 +787,20 @@ public static void unzip(String zipFile, String targetFolder, boolean overwrite,
...
+     * 
+     * @param targetDir target directory
+     * @param destPathStr destination file path
+     */
+    public static void validateDestPath(String targetDir, String destPathStr) throws IOException {
+        Path destPath = Paths.get(destPathStr);
+        Path destPathNormalized = destPath.normalize();
+        if (!destPathNormalized.toString().startsWith(targetDir + File.separator)) {
+            throw new IOException("Invalid output path: " + destPathStr);
+        }
```
coheigea: Please add some unit tests for this to check it's working as expected. Also I think there should be a check first that targetDir doesn't end with File.separator, otherwise remove it.

bhe-talendbj: Done.

PR #3037, opened 2020-02-04
bugfix(TUP-25990):Support MariaDB 10.X

hwang-talend:


main/plugins/org.talend.libraries.jdbc.mysql/plugin.xml
```diff
@@ -18,11 +18,11 @@
         </libraryNeeded>
         <libraryNeeded
               context="plugin:org.talend.libraries.jdbc.mysql"
-              id="mariadb-java-client-1.1.7.jar"
+              id="mariadb-java-client-2.5.3.jar"
               language="java"
               message="Needed for Mysql jdbc plugin"
-              mvn_uri="mvn:org.talend.libraries/mariadb-java-client-1.1.7/6.0.0"
-              name="mariadb-java-client-1.1.7.jar"
+              mvn_uri="mvn:org.talend.libraries/mariadb-java-client-2.5.3/6.0.0"
```
wchen-talend: component use mvn:org.mariadb.jdbc/mariadb-java-client/2.5.3 and this uploaded to nexus with this url .
better use the same mvnurl

PR #3050, opened 2020-02-07
fix(TUP-25986)Can't import attached job

jding-tlnd:
https://jira.talendforge.org/browse/TUP-25986
remove pigudf related


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/ui/actions/CopyObjectAction.java
```diff
@@ -471,24 +470,7 @@ private void synDuplicatedRoutine(RoutineItem item, String oldLabel) {
...
-            } else {
-                codeGenService.createRoutineSynchronizer().renameRoutineClass(item);
-            }
-            try {
-                codeGenService.createRoutineSynchronizer().syncRoutine(item, true);
-            } catch (SystemException e) {
-                ExceptionHandler.process(e);
-            }
-        }
-    }
```
wchen-talend: I think routine related code still need to keep in this function

PR #3051, opened 2020-02-07
bugfix(TUP-26031):Cannot connect successfully for mongDB which is

hwang-talend:
created from 721


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/service/INOSQLService.java
```diff
@@ -30,5 +30,5 @@
 
     public String getMongoDBReplicaSets(Connection connection);
 
-
+    public boolean updateNoSqlConnection(Connection connection)  throws Exception;
```
bhe-talendbj: which class will implement this method? Looks like need to update the NOSQLService.class as well?

PR #3057, opened 2020-02-10
bugfix(TUP-26073) Switch daikon to release version

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-26073


nrousseau:
already updated by wangwei

PR #3065, opened 2020-02-10
fix(TESB-27788): Fix issue by packing takokit components along with t…

vdrokov:
…he microservice jar.

1) Place takokit jars into /BOOT-INF/classpath/TALEND-INF/repository/<mvn path> directory
2) Place plugins description into /TALEND-INF/plugins.properties file.

### DIFF
--- a/main/plugins/org.talend.designer.maven/resources/templates/maven_user_settings.xml
+++ b/main/plugins/org.talend.designer.maven/resources/templates/maven_user_settings.xml
@@ -3,6 +3,15 @@
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 http://maven.apache.org/xsd/settings-1.0.0.xsd">
 
+	<mirrors>
+		<mirror>
+			<id>official</id>
+			<name>repo.maven.apache.org</name>
+			<url>https://repo.maven.apache.org/maven2</url>
+			<mirrorOf>central</mirrorOf>
+		</mirror>
+	</mirrors>
+
 	<!-- localRepository | The path to the local repository maven will use to 
 		store artifacts. | | Default: ${user.home}/.m2/repository -->
 	<localRepository>${user.home}/.m2/repository</localRepository>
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.designer.maven/resources/templates/maven_user_settings.xml
```diff
@@ -3,6 +3,15 @@
...
 
+	<mirrors>
+		<mirror>
+			<id>official</id>
+			<name>repo.maven.apache.org</name>
+			<url>https://repo.maven.apache.org/maven2</url>
+			<mirrorOf>central</mirrorOf>
+		</mirror>
+	</mirrors>
+
```
yanbasic: This part is not related to this issue, please do not commit it.

main/plugins/org.talend.libraries.jackson/META-INF/MANIFEST.MF
```diff
@@ -35,7 +35,7 @@ Require-Bundle: org.apache.commons.lang,
  org.apache.commons.beanutils,
  jackson-core-asl;bundle-version="1.9.13";visibility:=reexport,
  jackson-mapper-asl;bundle-version="1.9.13";visibility:=reexport,
- com.fasterxml.jackson.core.jackson-annotations;bundle-version="2.9.0";visibility:=reexport,
+ com.fasterxml.jackson.core.jackson-annotations;bundle-version="2.9.5";visibility:=reexport,
```
yanbasic: It's not idealy to change the jackson library in the maintenance branch, also some potential side effects need to check with other plugins, and we also need to provide this version into repository in studio before to upgraeding the versions.

main/plugins/org.talend.designer.maven/resources/templates/maven_user_settings.xml
vdrokov: Will not work without it

vdrokov: Because we need to build a microservice and maven issue will not allow to make a build

main/plugins/org.talend.libraries.jackson/META-INF/MANIFEST.MF
vdrokov: Ideally there will not be any side effects because we align the version of the library with version of the other libraries of the same family, this will make it work more safe

dsergent:
Will be handle in scope of a new feature project for supporting TaCoKit components in OSGi Builds.

PR #3067, opened 2020-02-11
fix(TDI-43626): Updated type generation for MySQL dynamic schema

OleksiiNimych:
**What is the current behavior?** (You can also link to an open issue here)

https://jira.talendforge.org/browse/TDI-43626

In job tSalesforceInput->tMysqlOutput using dynamic schema, occurs an SQLSyntaxErrorException. Discovered that during schema generation length for boolean fields does not exist. During SQL generation, a default value (-1) is taken for BIT type. Negative length for BIT type leads to an SQLSyntaxErrorException. Also occurs SQLSyntaxErrorException when schema has string fields that exceeds max length of VARCHAR type in MySQL.

**What is the new behavior?**

Now when creating queries to the MySQL database, the length for BIT types is not set. And for strings that exceeds max length of VARCHAR uses TEXT type.

**Please check if the PR fulfills these requirements**

- [x] The commit message follows Talend standard
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features) ?
- [ ] The code coverage on new code >75%
- [x] The new code does not introduce new technical issues (sonar / eslint)

**What kind of change does this PR introduce?**

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build / CI related changes
- [ ] Other... Please describe:

**Does this PR introduce a breaking change?**

- [ ] Yes
- [x] No

If this PR contains a breaking change, please describe the impact and migration path for existing applications: ...

**Other information**:


main/plugins/org.talend.core.runtime/mappings/mapping_Mysql.xml
```diff
@@ -6,7 +6,7 @@
 			<dbType type="BIGINT" ignorePre="true"/>
 			<dbType type="BIGINT UNSIGNED" ignoreLen="true" ignorePre="true" />
 			<dbType type="BINARY" ignorePre="true"/>
-			<dbType type="BIT" ignorePre="true" />
+			<dbType type="BIT" ignoreLen="true" ignorePre="true" />
```
gonchar-ivan: I'm not sure this fix is correct
According to the doc https://dev.mysql.com/doc/refman/8.0/en/bit-type.html 
BIT type accepts length value. 
"create table t3 (c1 BIT(62) NOT NULL);" is valis SQL query.

The problem is that MySQL generates query, in which it uses -1 for Bit types:
```
`Id` VARCHAR(18) ,`IsDeleted` BIT(-1) ,
```
Possible solution may be to save length as 1 for boolean values which come from Salesforce.
Another idea is to check SQL query which generated when MysqlInput->MysqlOutput job is executed

gonchar-ivan: UPD:
save length as 1 on Salesforce side is not good solution. It will work with Mysql, but may break if some other db is used as output.
It is better to try to fix it on Mysql side. If length is not specified, then generate column as BIT instead of BIT(-1)

mbasiuk-talend:
Close PR as not required for fixing original issue.

mbasiuk-talend:
It was requested from @OleksiiNimych

PR #3095, opened 2020-02-18
bugfix(TDQ-18070): Improve file path validation

bhe-talendbj:
https://jira.talendforge.org/browse/TDQ-18070

### DIFF
--- a/main/plugins/org.talend.utils/src/main/java/org/talend/utils/io/FilesUtils.java
+++ b/main/plugins/org.talend.utils/src/main/java/org/talend/utils/io/FilesUtils.java
@@ -795,11 +795,9 @@ public final class FilesUtils {
      */
     public static void validateDestPath(String targetDir, String destPathStr) throws IOException {
         if (targetDir != null && !targetDir.isEmpty() && destPathStr != null && !destPathStr.isEmpty()) {
-            Path destPathNormalized = Paths.get(destPathStr).normalize();
-            if (!targetDir.endsWith(File.separator)) {
-                targetDir += File.separator;
-            }
-            if (!destPathNormalized.startsWith(targetDir)) {
+            Path targetPathNormalized = Paths.get(new File(targetDir).toPath().toString()).normalize();
+            Path destPathNormalized = Paths.get(new File(destPathStr).toPath().toString()).normalize();
+            if (!destPathNormalized.startsWith(targetPathNormalized)) {
                 throw new IOException("Invalid output path: " + destPathStr);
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.utils/src/main/java/org/talend/utils/io/FilesUtils.java
```diff
@@ -795,8 +795,8 @@ public static void unzip(String zipFile, String targetFolder, boolean overwrite,
      */
     public static void validateDestPath(String targetDir, String destPathStr) throws IOException {
         if (targetDir != null && !targetDir.isEmpty() && destPathStr != null && !destPathStr.isEmpty()) {
-            Path targetPathNormalized = Paths.get(targetDir).normalize();
-            Path destPathNormalized = Paths.get(destPathStr).normalize();
+            Path targetPathNormalized = Paths.get(new File(targetDir).toPath().toString()).normalize();
+            Path destPathNormalized = Paths.get(new File(destPathStr).toPath().toString()).normalize();
```
coheigea: I think it's probably just easier to work with the File objects here?
File target = new File(targetDir);
File dest = new File(destPathStr);
 if (!dest.getCanonicalPath().startsWith(target.getCanonicalPath())) { throw new ...}

bhe-talendbj: > I think it's probably just easier to work with the File objects here?
> File target = new File(targetDir);
> File dest = new File(destPathStr);
> if (!dest.getCanonicalPath().startsWith(target.getCanonicalPath())) { throw new ...}


If I make above suggested change, following test case will fail, because string startsWith "test", actually need to check startsWith "test/"

```
// negative test cases
        targetDir = "/c/test/";
        destPathStr = "/c/test/../test1/file1.txt";

        try {
            FilesUtils.validateDestPath(targetDir, destPathStr);
            fail("validation should not pass");
        } catch (IOException e) {
            assertNotNull(e);
        }
```

So in sum, need further change as follows

```
 String target = new File(targetDir).getCanonicalPath();
String dest = new File(destPathStr)getCanonicalPath();
if (!target.endsWith(File.separator)){
   target+=File.separator;
}
if (!dest.startsWith(target)) { throw new ...}
```

So would you prefer make above changes?

coheigea: In the case, the original is fine I think.

PR #3105, opened 2020-02-25
fix(TUP-26165)jar referenced in bat or sh classpath missing in build ZIP

jding-tlnd:
file folder lib
https://jira.talendforge.org/browse/TUP-26165

### DIFF
--- a/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
+++ b/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
@@ -730,9 +730,16 @@ public class CreateMavenJobPom extends AbstractMavenProcessorPom {
         }
         
         try {
-            Set<Dependency> codesDependencies = PomUtil.getCodesDependencies(ERepositoryObjectType.ROUTINES);
+            Set<Dependency> codesDependencies = new HashSet<Dependency>();
+            for (ERepositoryObjectType type : ERepositoryObjectType.getAllTypesOfCodes()) {
+                codesDependencies.addAll(PomUtil.getCodesDependencies(type));
+            }
             for (Dependency codeDependency : codesDependencies) {
-                addToDuplicateLibs(duplicateLibs, codeDependency);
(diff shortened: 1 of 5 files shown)


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
```diff
@@ -725,6 +729,18 @@ protected void updateDependencySet(IFile assemblyFile) {
                 }
             }
         }
+        
+        try {
+            if (!routinesList.isEmpty()) {
```
zwxue: condition is not accurate here, some irrelevant user routine may have extra dependencies, should check dependencies in routine, beans, pigudf(<=7.2.1) for any case

main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
```diff
@@ -815,7 +831,17 @@ private void addToDuplicateLibs(Map<String, Set<Dependency>> map, Dependency dep
             Set<Dependency> set = new HashSet<>();
             map.put(coordinate, set);
         }
-        map.get(coordinate).add(dependency);
+        boolean exist = false;
+        for (Dependency dep : map.get(coordinate)) {
```
zwxue: just overwrite hasCode and equals of SortableDependeny is ok, no need to compare here

main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
jding-tlnd: solved, thanks

main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/creator/CreateMavenJobPom.java
jding-tlnd: solved, thanks

PR #3118, opened 2020-03-05
fix(TUP-26329)impala metadata not in sync with the impala component

hzhao-talendbj:


hcyi:
1, The new add text should support context .
2, Check the class ComponentToRepositoryProperty , the same way  in class
RepositoryToComponentProperty.java 
3, It is not the same behavior in component tImpalaConnection / tImpalaInput/etc for the password

hzhao-talendbj:
1. add context part
2.add componentToRepository
3. checked with weimin , the password part is ok

PR #3128, opened 2020-03-10
TMDM-14505 MDM connection in metadata randomly failed with error 'Failed to connect to MDM, please change MDM settings'

sbliu:
https://jira.talendforge.org/browse/TMDM-14505.

Add studio token to created mdm metadata connection.

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/service/IMDMWebServiceHook.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/service/IMDMWebServiceHook.java
@@ -1,6 +1,6 @@
 // ============================================================================
 //
-// Copyright (C) 2006-2019 Talend Inc. - www.talend.com
+// Copyright (C) 2006-2020 Talend Inc. - www.talend.com
 //
 // This source code is available under agreement available at
 // %InstallDIR%\features\org.talend.rcp.branding.%PRODUCTNAME%\%PRODUCTNAME%license.txt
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/service/IMDMWebServiceHook.java
```diff
@@ -0,0 +1,30 @@
+// ============================================================================
+//
+// Copyright (C) 2006-2019 Talend Inc. - www.talend.com
```
lpw82gf: Should be ` 2006-2020` now

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/service/IMDMWebServiceHook.java
```diff
@@ -0,0 +1,30 @@
...
+//
+// ============================================================================
+package org.talend.core.service;
+
+import java.util.Map;
+
+import org.talend.core.IService;
+
+/**
+ * created by HHB on 2013-12-23 Detailled comment
```
lpw82gf: by `HHB` ?

test/plugins/org.talend.metadata.managment.test/src/org/talend/metadata/managment/mdm/S60MdmConnectionHelperTest.java
```diff
@@ -0,0 +1,79 @@
+// ============================================================================
+//
+// Copyright (C) 2006-2019 Talend Inc. - www.talend.com
```
lpw82gf: to `2006-2020`

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/service/IMDMWebServiceHook.java
cmeng-talend: seems it is still not updated yet

PR #3139, opened 2020-03-16
fix(TUP-25863): add databricks cloud provider choice in wizard

jiezhang-tlnd:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/database/conn/ConnParameterKeys.java
```diff
@@ -304,6 +304,8 @@
     /**DataBricks*/
     public static final String CONN_PARA_KEY_DATABRICKS_ENDPOINT="CONN_PARA_KEY_DATABRICKS_ENDPOINT";
 
+    public static final String CONN_PARA_KEY_DATABRICKS_CLOUD = "CONN_PARA_KEY_DATABRICKS_CLOUD";
```
cmeng-talend: It is better to keep the meaning of the variable same with component: CONN_PARA_KEY_DATABRICKS_CLOUD_PROVIDER

PR #3142, opened 2020-03-18
fix(TUP-23956)CreateTable component should not to use JDBC connection

jiezhang-tlnd:


jding-tlnd:
Tested on this branch, exist a issue, please check:
Create Oracle with SID / PostgreSQL database under Metadata, create a tCreateTable component in a job, choose Oracle / PostgreSQL DBtype, choose repository Property type, click button to choose repository database from popup dialog, but we can see there is empty under DB connection node, please also check others DB.

PR #3157, opened 2020-03-24
TUP-26591 - Harden deserialization for LocalLibraryManager

coheigea:


coheigea:
Note: I can't compile this, so please test it properly before merging.

PR #3165, opened 2020-03-26
fix(TUP-26596)Proxy of libraries not working when studio have no

hzhao-talendbj:
internet access


hzhao-talendbj:
> Maybe we can use `TalendLibsServerManager.getInstance().getTalentArtifactServer().getServer()` dirrectly?

Remove the util file and replace with TalendLibsServerManager.getInstance().getTalentArtifactServer().getServer() ,  please review again

PR #3192, opened 2020-04-07
Kjwang/feat tup 26288 enhance context propagation

kjwang-talend:
Fix:TUP-26695 Enhance context propagation over reference project - Migration
https://jira.talendforge.org/browse/TUP-26695


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/update/RepositoryContextUpdateService.java
```diff
@@ -0,0 +1,400 @@
...
+                            sapProperty.setValue(newValue);
+                            isModified = true;
+                        }
+                    }
+                }
+            }
+        }
+        return isModified;
+    }
+
```
wchen-talend, line 330: Can you please create a single function for each connection ? painful to read a long function like this

PR #3196, opened 2020-04-08
fix(TBD-10269): method getDate can work with java.sql.Timestamp/Date

AlixMetivier:


main/plugins/org.talend.libraries.mdm.webservice.ce/.project
```diff
@@ -0,0 +1,28 @@
...
+			<name>org.eclipse.pde.SchemaBuilder</name>
+			<arguments>
+			</arguments>
+		</buildCommand>
+	</buildSpec>
+	<natures>
+		<nature>org.eclipse.pde.PluginNature</nature>
+		<nature>org.eclipse.jdt.core.javanature</nature>
+	</natures>
+</projectDescription>
```
talend-jphautin, line 28: what is this ?

AlixMetivier: thx I did not see it, the project was deleted but i still had it locally and did not notice reuploading this ! removing it

PR #3200, opened 2020-04-09
TUP-26813 - Update BouncyCastle to 1.65

coheigea:


bhe-talendbj (commented):
Could you please help close this PR, this is already done. please refer to https://github.com/Talend/studio-tools-se/blob/master/studio-utils/pom.xml

coheigea:
@bhe-talendbj - We still have an BC dependency in main/plugins/org.talend.utils/pom.xml - should I update the PR to change this to 1.65 as per the changes in studio-tools-se?

nrousseau:
close as all the dependencies changed on org.talend.utils regarding jobserver dependencies

PR #3214, opened 2020-04-15
Hwang/tup 26870

hwang-talend:


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/RepositoryNodeSortUtil.java
```diff
@@ -0,0 +1,56 @@
...
+        
+        Collections.sort(temp, new Comparator<IRepositoryViewObject>() {
+
+            @Override
+            public int compare(IRepositoryViewObject o1, IRepositoryViewObject o2) {
+                String version1 = o1.getVersion();
+                String version2 = o2.getVersion();
+                if(version1 != null && version2 != null) {
+                	try {
+                		float v1 = Float.parseFloat(version1);
```
kjwang-talend, line 40: How to process the version like 1.11? I think it have problem. Can you use org.talend.commons.utils.Version, it implements the interface Comparable.

PR #3216, opened 2020-04-15
bugfix(TUP-26875):when import a job with many versions,the order of the

hwang-talend:
version is not right


main/plugins/org.talend.repository.items.importexport.ui/src/main/java/org/talend/repository/items/importexport/ui/wizard/imports/providers/ImportItemsViewerSorter.java
```diff
@@ -62,6 +63,43 @@ public int compare(Viewer viewer, Object o1, Object o2) {
...
+                	
+                	if (version1 == null) {
+                        return -1;
+                    }
+                    if (version2 == null) {
+                        return 1;
+                    }
+                    try {
+                		float v1 = Float.parseFloat(version1);
+                        float v2 = Float.parseFloat(version2);
```
kjwang-talend, line 91: How to process the version like 1.11? I think it have problem. Can you use org.talend.commons.utils.Version, it implements the interface Comparable.

PR #3245, opened 2020-04-27
fix(TDI-44093): Add method that checks if string wrapped by quotes

OleksiiNimych:
**What is the current behavior?** (You can also link to an open issue here)
https://jira.talendforge.org/browse/TDI-44093
Syntax error when use query for tCosmosDBSQLAPIInput. Logic remove and add quotes by default.

**What is the new behavior?**
Currently strings with quoted values process as expected.
New method check is it necessary quote string

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/TalendQuoteUtils.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/TalendQuoteUtils.java
@@ -281,6 +281,20 @@ public final class TalendQuoteUtils {
         return text;
     }
 
+    public static boolean checkTextQuotation(String text) {
+        if (text == null || text.length() == 0) {
+            return false;
+        } else {
+            String begin = text.substring(0, 1);
+            String end = text.substring(text.length() - 1, text.length());
+            if (QUOTATION_MARK.equals(begin) && QUOTATION_MARK.equals(end)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+    }
+
     /**
      *
      * ggu Comment method "addQuotesForSQLString".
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/TalendQuoteUtils.java
```diff
@@ -281,6 +281,20 @@ public static String removeQuotes(String text, String quotation) {
...
+            String begin = text.substring(0, 1);
+            String end = text.substring(text.length() - 1, text.length());
+            if (QUOTATION_MARK.equals(begin) && QUOTATION_MARK.equals(end)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
+    }
+
```
gonchar-ivan, line 297: String class has following nice methods endsWith() and startsWith() and other ;-)
I would refactor this method in following way:
```
if (text == null) {
    return false;
}
text = text.trim();
if (text.isEmpty()) {
    return false;
}
return text.startsWith(QUOTATION_MARK) && text.endsWith(QUOTATION_MARK)
```

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/TalendQuoteUtils.java
```diff
@@ -281,6 +281,20 @@ public static String removeQuotes(String text, String quotation) {
         return text;
     }
 
+    public static boolean checkTextQuotation(String text) {
```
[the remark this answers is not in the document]
gonchar-ivan: actually the name could even be isEnclosed(String text)
(there is some recommendation to to not have more than 3 words in method name)

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/utils/TalendQuoteUtils.java
cmeng-talend: May have bug if inputting text: `"` (a single char `"`), please replace the `text.isEmpty()` to `text.length() < 2`  ;-)

PR #3246, opened 2020-04-28
bugfix(TUP-26956):Update Studio top bar to reflect monthly deliveries

hwang-talend:
for better understanding


main/plugins/org.talend.repository.localprovider/src/main/java/org/talend/repository/localprovider/model/LocalRepositoryFactory.java
```diff
@@ -3275,10 +3276,23 @@ public void unloadUnlockedResources() {
...
     public void beforeLogon(Project project) throws PersistenceException, LoginException {
+    	String productVersion = VersionUtils.getDisplayVersion();
+    	
+    	if (GlobalServiceRegister.getDefault().isServiceRegistered(IInstalledPatchService.class)) {
+    		IInstalledPatchService pachService = (IInstalledPatchService) GlobalServiceRegister
+                    .getDefault().getService(IInstalledPatchService.class);
+            if (pachService != null && pachService.isMonthlyPatch()) {
+                String patchVersion = pachService.getLatestInstalledVersion();
+                if(patchVersion != null) {
+                	System.setProperty(VersionUtils.STUDIO_VERSION_PROP, patchVersion);
```
wchen-talend, line 3287: STUDIO_VERSION_PROP is used for many place like talend.project/ .properties , easy to cause bug if change like this .

wchen-talend: I think the most simple way is to add a function in ICoreTisService like getTopBarDisplayVersion.
then update the dispaly version https://github.com/Talend/tcommon-studio-se/blob/kjwang/feat_TUP-26288_Enhance_context_propagation/main/plugins/org.talend.rcp/src/main/java/org/talend/rcp/intro/ApplicationWorkbenchWindowAdvisor.java#L187

PR #3249, opened 2020-04-29
fix(TUP-26639):Talend Cloud studio Issue -Changes to Context Parameter

jding-tlnd:
Names is not reflected on the Connection for Snowflake
https://jira.talendforge.org/browse/TUP-26639

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/update/RepositoryUpdateManager.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/update/RepositoryUpdateManager.java
@@ -1136,6 +1136,9 @@ public abstract class RepositoryUpdateManager {
                             }
                         }
                     }
+                    if (foundCompProperties) {
+                        break;
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/update/RepositoryUpdateManager.java
```diff
@@ -1109,10 +1109,39 @@ private void updateConnection(ContextItem citem) throws PersistenceException {
...
+                            if (conn.isContextMode()) {
+                                ContextItem contextItem = ContextUtils.getContextItemById2(conn.getContextId());
+                                if (contextItem == null) {
+                                    continue;
+                                }
+                                if (citem == contextItem) {
+                                    String compProperties = conn.getCompProperties();
+                                    if (StringUtils.isNotBlank(compProperties)) {
+                                        foundCompProperties = true;
+                                        dbConnectionItem = (ConnectionItem) item;
```
cmeng-talend, line 1132: Please break the whole loop after found the item

jding-tlnd: Modified, thanks

PR #3287, opened 2020-05-14
fix(TBD-10269): force return

AlixMetivier:


main/plugins/org.talend.librariesmanager/resources/java/routines/system/JDBCUtil.java
```diff
@@ -71,12 +71,14 @@ public static Date getDate(ResultSet rs, int index) throws java.sql.SQLException
         try {
             if(rs.getTimestamp(index) != null) {
                 result = new Date(rs.getTimestamp(index).getTime());
+                return result;
```
hliashchuk, line 74: we can shrink it to `return new Date(rs.getTimestamp(index).getTime());` but as you wish

PR #3289, opened 2020-05-14
fix(TDI-44128): delete axis2 jars

Voievidko:
https://jira.talendforge.org/browse/TDI-44128

To fix CVE issue axis2 libs was updated from 1.5 to 1.6.3.

Updated libs will be uploaded to Nexus talend-update.

Merge this PR after uploading deps to Nexus.


Voievidko:
Deps were uploaded:
https://talend-update.talend.com/nexus/content/repositories/libraries/org/apache/axis2/

Voievidko:
1. We don't have tuj with axis2 deps (it used only in tMicrosoftCrmInput and tMicrosoftCrmOutput)
2. Basically this bug was on TUP side, but they defined that jar used only on TDI. So seems we are safe here.

PR #3290, opened 2020-05-14
TUP-27206 - Update Apache Ant to 1.9.15

coheigea:


nrousseau:
part in refactor now with: TUP-27438
not sure it's worth to update during the refactor here.

bhe-talendbj (commented):
Could you please help close this PR, this is already done. please refer to https://github.com/Talend/studio-tools-se/blob/master/studio-utils/pom.xml

coheigea:
@bhe-talendbj - We still have an Ant 1.9.12 dependency in main/plugins/org.talend.utils/pom.xml - should I update the PR to change this to 1.10.7 as per the changes in studio-tools-se?

nrousseau:
not needed anymore, as jobserver is not using anymore this jar

PR #3460, opened 2020-07-09
TUP-27849

hzhao-talendbj:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/database/conn/version/EDatabaseVersion4Drivers.java
```diff
@@ -107,7 +107,7 @@
 
     PLUSPSQL_PRIOR_TO_V9(new DbVersion4Drivers(EDatabaseTypeName.PLUSPSQL,
```
bhe-talendbj, line 108: 1 - According to https://jdbc.postgresql.org/download.html, 42.2.14 supports PostgreSQL 8.2 or newer and requires Java 6 or newer. Can we Remove postgresql-8.4-703.jdbc4.jar as well?

main/plugins/org.talend.libraries.jdbc.postgresql/build.properties
```diff
@@ -1,5 +1,4 @@
 output.. = bin/
 bin.includes = META-INF/,\
                .,\
-               lib/postgresql-8.4-703.jdbc4.jar,\
-               lib/postgresql-9.4-1201.jdbc41.jar
+               lib/postgresql-8.4-703.jdbc4.jar
```
bhe-talendbj, line 4: Looks like missing postgresql-42.2.14.jar ?

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/database/conn/version/EDatabaseVersion4Drivers.java
hzhao-talendbj: removed  postgresql-8.4-703.jdbc4.jar and relative codes.

main/plugins/org.talend.libraries.jdbc.postgresql/build.properties
hzhao-talendbj: added into build.properties

PR #3474, opened 2020-07-14
Fix:TUP-27879 Shared user can't write library index file library_data.index

kjwang-talend:
Fix:TUP-27879 Shared user can't write library index file library_data.index
https://jira.talendforge.org/browse/TUP-27879


bhe-talendbj:
LGTM

PR #3495, opened 2020-07-20
Enable precision check for Timestamp datatype.

dmytrogrygorenko:
Fix for https://jira.talendforge.org/browse/TDI-44457 issue.


dmytro-sylaiev:
Don't forget to run tujs to verify nothing is broken.

PR #3498, opened 2020-07-22
fix(TBD-10269): force return in case of bad job design (#3287)

AlixMetivier:


main/plugins/org.talend.librariesmanager/resources/java/routines/system/JDBCUtil.java
```diff
@@ -58,11 +58,22 @@ public static boolean getBoolean(ResultSet rs, int index) throws java.sql.SQLExc
...
-        
-        return null;
+        try {
+            if(rs.getDate(index) != null) {
+                result = new Date(rs.getDate(index).getTime());
+                return result;
+            }
+        } catch (java.sql.SQLException e) {
+        }
+        return result;
```
amedvedenko, line 76: it'ss just a cherry-pick, right?

AlixMetivier: to confess, since there were two commits, I've done it by hand, just needs to be the same than master :)

PR #3513, opened 2020-07-29
feat(TUP-23827):Runtime Lineage for studio.

hcyi:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/utils/NodeUtil.java
```diff
@@ -1232,4 +1233,13 @@ private static boolean isConnectorValid(INodeConnector mainConnector) {
 
         return true;
     }
+
+    public static boolean isJobUsingRuntimeLineage(IProcess process) {
+        RuntimeLineageManager runtimeLineageManager = new RuntimeLineageManager();
+        if (runtimeLineageManager.getDynamicFields().isEmpty()) {
+            runtimeLineageManager.load();
```
wchen-talend, line 1240: if isUseRuntimeLineageAll is true , no need to load it at all.

wchen-talend (approved):

PR #3574, opened 2020-08-21
bugfix(TUP-28308) build error and joblet maven xml file has incorrect values and Couldn't parse data error message

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-28308

### DIFF
--- a/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/AggregatorPomsHelper.java
+++ b/main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/AggregatorPomsHelper.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.designer.maven/src/main/java/org/talend/designer/maven/tools/AggregatorPomsHelper.java
```diff
@@ -368,6 +372,29 @@ private static boolean checkIfCanAddToParentModules(Property property, boolean c
                     && PomIdsHelper.getIfExcludeDeletedItems(property)) {
                 return false;
             }
+
+            if (Boolean.getBoolean(SYS_PROP_INCLUDE_REFERENCED_JOBLETS_ONLY)) {
```
nrousseau, line 376: How do you plan to use that?
as performance for testing only?

As in fact it should be the best thing to do by default even... only possible bug, is like: user have wrong index file right?

zwxue (approved):

PR #3589, opened 2020-08-27
fix(TUP-27268): Add isRequiredAlways for ILoginTask

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-27268


cmeng-talend (approved):
LGTM

PR #3596, opened 2020-08-28
TUP-27924 m2 jars in patch should be synchronized right after patch installed but not in logon phase

sbliu:
https://jira.talendforge.org/browse/TUP-27924


sbliu:
re-created a new clean PR.

PR #3598, opened 2020-08-30
fix(TUP-28487): fix performance issue

nrousseau:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/context/ContextUtils.java
```diff
@@ -507,7 +508,7 @@ public static Item getRepositoryContextItemById(String contextId) {
 
         IProxyRepositoryFactory factory = CoreRuntimePlugin.getInstance().getProxyRepositoryFactory();
         try {
-            final IRepositoryViewObject lastVersion = factory.getLastVersion(contextId);
+            final IRepositoryViewObject lastVersion = factory.getLastVersion(contextId, ERepositoryObjectType.CONTEXT);
             if (lastVersion != null) {
```
kjwang-talend, line 512: I think we can't modify the code like this. the contextId can be either joblet node or context node

PR #3600, opened 2020-08-31
feature(TUP-28334) Enhance logging of performance.

sbliu:
https://jira.talendforge.org/browse/TUP-28334

### DIFF
--- a/main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
+++ b/main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
@@ -72,6 +72,7 @@ import org.talend.commons.ui.runtime.exception.MessageBoxExceptionHandler;
 import org.talend.commons.utils.data.container.RootContainer;
 import org.talend.commons.utils.network.TalendProxySelector;
 import org.talend.commons.utils.time.TimeMeasure;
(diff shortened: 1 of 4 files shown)


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
```diff
@@ -2040,6 +2041,7 @@ public void logOnProject(Project project, IProgressMonitor monitor) throws Login
             TimeMeasure.measureActive = CommonsPlugin.isDebugMode();
 
             TimeMeasure.begin("logOnProject"); //$NON-NLS-1$
+            TimeMeasurePerformance.begin("logOnProject"); //$NON-NLS-1$
```
wchen-talend, line 2044: The new TimeMeasurePerformance will log the performance,  so TimeMeasure should be removed totally during logon ?

sbliu: I'm not sure to keep or not the TimeMeasure during logon, is there any possible still need TimeMeasure log during logon?

sbliu: As we discussed, will remove TimeMeasure log  from logon project part.

PR #3601, opened 2020-09-01
fix(TESB-29853): TESB-29819 Studio doesn't select model beans during …

vdrokov:
…exporting


andreasmattes (approved):
Fix is identical with the approved pull request 3512, here for the master branch.

andreasmattes:
Merge after successful validation with locally patched Studio.

PR #3665, opened 2020-09-21
fix(TDI-44906):fix the dynamic error for dbtype of 'money'.

vyu-talend:
https://jira.talendforge.org/browse/TDI-44906


pyzhou-talend (approved):
LGTM

PR #3686, opened 2020-09-28
feat(TUP-28640):Improve JDBC database support framework to load

jding-tlnd:
supported DBs and components automatically
https://jira.talendforge.org/browse/TUP-28640


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
```diff
@@ -2140,6 +2141,14 @@ public void logOnProject(Project project, IProgressMonitor monitor) throws Login
 
                 ProjectDataJsonProvider.checkAndRectifyRelationShipSetting(project.getEmfProject());
 
+                // load additional jdbc
```
wchen-talend, line 2144: why need to load the additional jdbc during the logon.
It is only needed in the database wizard and job.
Better load after logon to avoid increase time for logon studio.

jding-tlnd: can't do the load on load component since I found that the load component running in the background thread, so might meet that studio starts already but load component still running, at that time when we open the connection wizard, the Delta Lake dbType missing

PR #3692, opened 2020-10-05
fix(TDI-44962): fix stream closed exception

chmyga:
See https://jira.talendforge.org/browse/TDI-44962 + https://jira.talendforge.org/browse/TDI-44826
Recreating the SimpleCsvWriter as it was done in fix for https://jira.talendforge.org/browse/TDI-44826 and fixing the problem of stream closed by adding the SimpleCsvWriter to sharedWriterMap.


rlecomte-talend (approved):
LGTM

PR #3694, opened 2020-10-08
fix(TBD-11038): The distribution AMAZON_EMR with the version APACHE_2_4_0_EMR doesn't exist

cbadillo1603:
…_4_0_EMR doesn't exist


lbourgeois (approved):
LGTM

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/components/ComponentUtilities.java
```diff
@@ -81,7 +81,9 @@ public static void setNodeProperty(NodeType node, String name, List<ElementValue
 
     public static void setNodeValue(NodeType node, String name, String value) {
         ElementParameterType property = getNodeProperty(node, name);
-        property.setValue(value);
+        if (property != null) {
+            property.setValue(value);
+        }
```
amedvedenko, line 86: Well, it'll work but, as I understand, it will hide an error if there is. I think the issue is in other place. Agree?

cbadillo1603: Whenever there is a node that is not able to find the property requested. This will throw a NullPointerException I agree the issue will be in another place, so I will review thoroughly the main error why I had to add this snippet. However, I'm still hesitating whether to include this change or not.

amedvedenko (approved):

amedvedenko (approved):

PR #3730, opened 2020-10-16
add all jars in lib folder

hzhao-talendbj:


hzhao-talendbj:
close as it won't re-build M3,  will fix on master

PR #3827, opened 2020-11-11
fix(TUP-29165): Add cache for all of current project nodes

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-29165

### DIFF
--- a/main/plugins/org.talend.core.runtime/src/main/java/org/talend/repository/ProjectManager.java
+++ b/main/plugins/org.talend.core.runtime/src/main/java/org/talend/repository/ProjectManager.java
@@ -20,6 +20,9 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.WeakHashMap;
+import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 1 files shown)


main/plugins/org.talend.core.runtime/src/main/java/org/talend/repository/ProjectManager.java
```diff
@@ -451,9 +453,18 @@ public boolean isInCurrentMainProject(IRepositoryNode node) {
                     if (object == null) {
                         return true;
                     }
+
+                    if (object.getId() != null && cachedObjects.containsKey(object.getId())) {
```
wchen-talend, line 457: There is one extreme condition might not work: main and reference project contain some item with the same ID.
Normally we do not support it  but we can't guarantee customer do not have this kind of case .
After this commit , object with the same id in reference project might be editable.
Possible to store the key with project id+object id ?

bhe-talendbj: cache key changed to object itself

wchen-talend (approved):

PR #3862, opened 2020-11-19
fix(TUP-29360): fix switch log4j level problem

zwxue:


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/ModulesNeededProvider.java
```diff
@@ -648,13 +648,6 @@ private static Property findRoutinesPropery(String id, String name, List<IReposi
                             routinesName.add(routineName);
                         }
                     }
-                    if (routinesName.contains("BigDataUtil")) { //$NON-NLS-1$
```
zwxue, line 651: it got bugs this way, systemModules is the routine list cache which only init once, so switching log4j level won't work, so moved the code to CreateMavenRoutinePom

nrousseau (approved):

PR #3883, opened 2020-11-26
TESB-31234 context variables corresponding to TMC connections are not cleaned up in Studio during publish to Cloud

apoltavtsev:


dsergent (approved):
LGTM

PR #3890, opened 2020-12-01
TPSVC-15823 jobserver purge issues for maintenance/7.3

drdwo:
This PR is related to https://jira.talendforge.org/browse/TPSVC-15823
Remote Engine automatic purge doesn't work.

We experimented with different changes and @JuergenKindler s tests indicate that the solution implemented here solves the issue.


cmeng-talend (approved):
LGTM

PR #3891, opened 2020-12-01
TESB-31287:

vdrokov:
talend.version value is wrong in talend.properties in org.talend.commons.runtime_7.4.1.xxxxx.jar


dsergent (approved):
LGTM

PR #3892, opened 2020-12-02
Kjwang/fix tup 29525 update certificates 71

kjwang-talend:
TUP-29525 Update certificates for item / job signing on 7.0 / 7.1
https://jira.talendforge.org/browse/TUP-29525


cmeng-talend (approved):
LGTM

PR #3913, opened 2020-12-11
fix(TUP-29694):Studio fails to download jar with classifier in mvn uri

lbourgeois:
This change is aligned with ﻿https://maven.apache.org/pom.html.
Artifacts with and without classifier share the same general pom.


firefryy (approved):
lgtm

talend-jphautin (approved):

bhe-talendbj (approved):

PR #3925, opened 2020-12-16
fix(TBD-11813):Enhance migration PropertyComponentFilter with contains operator

lbourgeois:


main/plugins/org.talend.core/src/main/java/org/talend/core/model/migration/AbstractSparkJobMigrationTask.java
```diff
@@ -0,0 +1,33 @@
...
+ * @author lbourgeois
+ *
+ */
+public abstract class AbstractSparkJobMigrationTask extends AbstractJobMigrationTask {
+
+	@Override
+    public List<ERepositoryObjectType> getTypes() {
+	    // Spark jobs are stored under process_mr and process_storm folders
+        return Arrays.asList(ERepositoryObjectType.PROCESS_MR, ERepositoryObjectType.PROCESS_STORM);
+    }
```
kjwang-talend, line 32: Here should contains all the process type which may contains CDP DISTRIBUTION. e.g SPARK_JOBLET,SPARK_STREAMING_JOBLET,TEST_CONTAINER(only spark process related) Please ref to:
    public List<ERepositoryObjectType> getTypes() {
        if (ERepositoryObjectType.PROCESS_STORM != null) {
            // test if spark streaming is available (not the case with TBD license)
            return Arrays.asList(ERepositoryObjectType.PROCESS_MR, ERepositoryObjectType.PROCESS_STORM,ERepositoryObjectType.SPARK_JOBLET,ERepositoryObjectType.SPARK_STREAMING_JOBLET,ERepositoryObjectType.TEST_CONTAINER);
        } else {
            return Arrays.asList(ERepositoryObjectType.PROCESS_MR, ,ERepositoryObjectType.SPARK_JOBLET, ERepositoryObjectType.TEST_CONTAINER);
        }
    }

lbourgeois: Thanks, I will modify.

kjwang-talend (approved):

PR #3934, opened 2020-12-22
TESB-31444:[7.3.1] class org.apache.cxf.headers.Header cannot be cast to class org.apache.cxf.headers.Header

csuntalendbj:


dsergent (approved):
LGTM

PR #3938, opened 2020-12-22
Fix TUP-29744 The parameter value of jobType is wrong for new created testcase.

kjwang-talend:
Fix TUP-29744 The parameter value of jobType is wrong for new created testcase.
https://jira.talendforge.org/browse/TUP-29744

### DIFF
--- a/main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/ConvertJobsUtil.java
+++ b/main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/ConvertJobsUtil.java
@@ -677,6 +677,16 @@ public class ConvertJobsUtil {
         final Path path = new Path(pathName.toString());
         return path;
     }
+    
+    public static String getTestCaseJobTypeByPath(String itemPath) {
+        if (itemPath.startsWith(JobType.BIGDATASTREAMING.repositoryObjectType.getFolder())) {
+            return ConvertJobsUtil.JobType.BIGDATASTREAMING.getDisplayName();
+        } else if (itemPath.startsWith(JobType.BIGDATABATCH.repositoryObjectType.getFolder())) {
+            return JobType.BIGDATABATCH.getDisplayName();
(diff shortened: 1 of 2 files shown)


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/ConvertJobsUtil.java
```diff
@@ -677,6 +677,16 @@ public static IPath getTestCasePath(Item newItem, String jobTypeValue) {
         final Path path = new Path(pathName.toString());
         return path;
     }
+    
+    public static String getTestCaseJobTypeByPath(String itemPath) {
+        if (itemPath.startsWith(JobType.BIGDATASTREAMING.repositoryObjectType.getFolder())) {
```
cmeng-talend, line 682: need to avoid npe of RepositoryObjectType, since it can be null in some DI license

main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/utils/ConvertJobsUtil.java
```diff
@@ -677,6 +677,16 @@ public static IPath getTestCasePath(Item newItem, String jobTypeValue) {
         final Path path = new Path(pathName.toString());
         return path;
     }
+    
+    public static String getTestCaseJobTypeByPath(String itemPath) {
+        if (itemPath.startsWith(JobType.BIGDATASTREAMING.repositoryObjectType.getFolder())) {
+            return ConvertJobsUtil.JobType.BIGDATASTREAMING.getDisplayName();
+        } else if (itemPath.startsWith(JobType.BIGDATABATCH.repositoryObjectType.getFolder())) {
```
cmeng-talend, line 684: need to avoid npe of RepositoryObjectType, since it can be null in some DI license

cmeng-talend (changes requested):

PR #3947, opened 2020-12-28
Revert the commit for TESB-31444

csuntalendbj:


hzhao-talendbj (approved):

hzhao-talendbj:
will be fix in https://jira.talendforge.org/browse/TUP-29826  so revert this one

PR #3956, opened 2020-12-30
Kjwang/fix tup 29744 job type is wrong m

kjwang-talend:
Fix TUP-29744 The parameter value of jobType is wrong for new created testcase.
https://jira.talendforge.org/browse/TUP-29744


cmeng-talend (approved):
LGTM

PR #3960, opened 2021-01-04
feat(TUP-29651): [Research] Components loading improvement

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-29651


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/LocalLibraryManager.java
```diff
@@ -1265,16 +1263,10 @@ public void createModulesIndexFromComponentAndExtension(IProgressMonitor... moni
...
-                    modules.addAll(component.getModulesNeeded());
-                } catch (Exception e) {
-                    ExceptionHandler.process(e);
-                    continue;
-                }
-            }
-        }
+
+        // we need to invoke this, because some module needed is defined inside component plugin.xml
+        modules.addAll(ModulesNeededProvider.getAllManagedModules());
```
wchen-talend, line 1268: Modules defined in plugin.xml will be include in calculateModulesIndexFromExtension(.....) function .
ModulesNeededProvider.getAllManagedModules() includes custom maven url ,it will cause issue if put into default index.
It also might cause other potential issues.

PR #4005, opened 2021-01-20
fix(TUP-29586) Change repository context variable type would pop up job update dialog.

sbliu:
https://jira.talendforge.org/browse/TUP-29586

after save changed context , popup dialog to update job.


sbliu:
Just remove the if clause, because it always should be true.

jding-tlnd (approved):

PR #4065, opened 2021-02-19
item(TUP-30180): Studio slowness on merge operation

cmeng-talend:
https://jira.talendforge.org/browse/TUP-30180


cmeng-talend:
wrong PR

PR #4082, opened 2021-02-25
fix(TBD-11964): Different time zone for Standard jobs and BD Batch

lbourgeois:


main/plugins/org.talend.librariesmanager/resources/java/routines/TalendDate.java
```diff
@@ -1221,7 +1241,7 @@ public static void test_formatDate() {
 
             @Override
             public void run() {
-                Calendar calendar = GregorianCalendar.getInstance();
+                Calendar calendar = Calendar.getInstance();
```
AlixMetivier, line 1244: curiosity question, whats the difference between GregorianCalendar & Calendar ?

AlixMetivier (approved):

main/plugins/org.talend.librariesmanager/resources/java/routines/TalendDate.java
lbourgeois: Honnestly this change was done automatically by Eclipse because getInstance is a member of Calendar. From javadoc   _GregorianCalendar is a concrete subclass of Calendar and provides the standard calendar system used by most of the world._

hliashchuk (approved):

PR #4121, opened 2021-03-15
fix(TUP-30277)Fail to download jar in remote project when tac set a

jiezhang-tlnd:
wrong nexus url.
https://jira.talendforge.org/browse/TUP-30277


wchen-talend:
Improve the connection check is a good point but still need to check if custom nexus is available or not before deploy the jar.

cmeng-talend (approved):

PR #4134, opened 2021-03-22
TESB-32452

vdrokov:
After applying the latest patch, trunjob component fails with "java.io.IOException: Cannot run program "java.exe": error=2, No such file or directory"


apoltavtsev (approved):
OK

PR #4151, opened 2021-03-28
fix(TUP-30849):Improve build Job performance

jding-tlnd:
https://jira.talendforge.org/browse/TUP-30849

Signed-off-by: jding-tlnd <jding@talend.com>


main/plugins/org.talend.core/src/main/java/org/talend/designer/runprocess/ProcessorUtilities.java
```diff
@@ -436,6 +434,91 @@ private static boolean isFileExist(IFile codeFile) {
         return false;
     }
 
+    private static IProcess getProcessFromCaches(String jobinfoId, String jobinfoVersion) {
```
zwxue, line 437: need to be careful of process cache, better to let QA have a test for TUP-20979

zwxue (approved):

PR #4213, opened 2021-04-16
feat(TUP-30846) Unify studio version information & patch info in studio logs.

sbliu:
https://jira.talendforge.org/browse/TUP-30846

### DIFF
--- a/main/plugins/org.talend.platform.logging/src/org/talend/platform/logging/Activator.java
+++ b/main/plugins/org.talend.platform.logging/src/org/talend/platform/logging/Activator.java
@@ -108,10 +108,8 @@ public class Activator extends AbstractUIPlugin {
                 //
             }
 
-            if (StringUtils.isEmpty(version)) {
-                if (version == null || "".equals(version.trim())) { //$NON-NLS-1$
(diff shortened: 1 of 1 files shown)


jding-tlnd (approved):

main/plugins/org.talend.platform.logging/src/org/talend/platform/logging/Activator.java
```diff
@@ -64,9 +72,37 @@ public static Activator getDefault() {
     }
 
     public static String getVersion() {
-        String version = System.getProperty("talend.studio.version"); //$NON-NLS-1$
-        if (version == null || "".equals(version.trim())) { //$NON-NLS-1$
-            version = (String) getDefault().getBundle().getHeaders().get(org.osgi.framework.Constants.BUNDLE_VERSION);
+        String version = null;
```
cmeng-talend, line 75: If commons.runtime is never patched, the bundle version won't be the studio patch version.
So it's better to make reading from commons.runtime as the last choice.
And please also consider performance here, no sure whether this method will be called frequntly

cmeng-talend (changes requested):

PR #4377, opened 2021-06-15
chore(TUP-31163): upgrade daikon to 0.31.12

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-31163

### DIFF
--- a/main/plugins/org.talend.migrationTool/src/org/talend/migrationtool/MigrationToolService.java
+++ b/main/plugins/org.talend.migrationTool/src/org/talend/migrationtool/MigrationToolService.java
@@ -277,6 +277,8 @@ public class MigrationToolService implements IMigrationToolService {
         // force to re-generate all job poms
         MigrationUtil.removeMigrationTaskById(done, "org.talend.repository.model.migration.GenerateJobPomMigrationTask");
 
+        MigrationUtil.removeMigrationTaskById(done, "org.talend.repository.model.migration.UpdateDaikonCryptoUtilsMigrationTask");
+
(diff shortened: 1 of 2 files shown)


kjwang-talend (approved):

main/plugins/org.talend.migrationTool/src/org/talend/migrationtool/MigrationToolService.java
```diff
@@ -277,6 +277,8 @@ private void delateExecuteMigrationTasksForLogon(final Project project, final bo
         // force to re-generate all job poms
         MigrationUtil.removeMigrationTaskById(done, "org.talend.repository.model.migration.GenerateJobPomMigrationTask");
 
+        MigrationUtil.removeMigrationTaskById(done, "org.talend.repository.model.migration.UpdateDaikonCryptoUtilsMigrationTask");
```
nrousseau, line 280: why put here...?
This means that if there is any migration task to be executed.. it will force to update daikon crypto.
But means if nothing to do (which is maybe the case between R2021-06 / R2021-07.. it won't update crypto..?)

bhe-talendbj: Yes, this will be executed always. however according to https://github.com/Talend/tdi-studio-se/pull/6330/files, in this migration task, it will check whether the daikon version is <0.31.12, if so, the task will be executed and poms will be regenerated, otherwise this taks does nothing.

kjwang-talend (approved):

PR #4466, opened 2021-07-27
fix(TUP-32333):Not able to access a repository resource file in TMC

jding-tlnd:
https://jira.talendforge.org/browse/TUP-32333


cmeng-talend (approved):
LGTM

PR #4492, opened 2021-08-04
fix(TUP-32428) upgrade plexus-io to fix build job problem.

sbliu:
https://jira.talendforge.org/browse/TUP-32428


sbliu:
patch of TUP-31799 is ok, only to fix original studio build problem in junit test, so do not need such change anymore.

PR #4542, opened 2021-08-26
feat(TUP-32281):Provide a project analysis report

jding-tlnd:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/analysistask/ItemAnalysisReportManager.java
```diff
@@ -0,0 +1,249 @@
...
+    private AtomicBoolean inGenerating = new AtomicBoolean(false);
+
+    public List<AnalysisReportRecorder> executeAnalysisTask(Project project) {
+        IRepositoryService service = (IRepositoryService) GlobalServiceRegister.getDefault().getService(IRepositoryService.class);
+        IProxyRepositoryFactory repFactory = service.getProxyRepositoryFactory();
+        List<AnalysisReportRecorder> analysisResultList = new ArrayList<AnalysisReportRecorder>();
+        List<IItemAnalysisTask> analysisTasks = ItemAnalysisTaskRegistryReader.getInstance().getItemAnalysisTasks();
+        ERepositoryObjectType[] types = (ERepositoryObjectType[]) ERepositoryObjectType.values();
+        try {
+            for (ERepositoryObjectType type : types) {
```
wchen-talend, line 80: Different from migration ,I think we do not need to loop all repository types.
Maybe currently we only added 3 analysis task for jobs , but it will still loop for all repository type.
I think we can define the repository types from extension/interface for analysis task and only check needed types.

jding-tlnd: modified, thanks

wchen-talend (dismissed):

kjwang-talend (dismissed):

kjwang-talend (approved):

PR #4555, opened 2021-09-01
feat(TBD-12880):Dynamic schema in Spark batch components

lbourgeois:


test/plugins/org.talend.core.runtime.test/src/org/talend/core/model/metadata/MetadataTalendTypeFilterTest.java
```diff
@@ -90,7 +123,20 @@ public void filterStringTest() {
...
+    }
 
+    @Test
+    public void sparkDynamicTypeTest() {
+        String[] types = new String[] { INTEGER, DOCUMENT, STRING, OBJECT, LIST, DOUBLE, SHORT, DYNAMIC, VECTOR };
+        IComponent component = mock(IComponent.class);
+        when(node.getComponent()).thenReturn(component);
+        for (String componentName: SparkBatchMetadataTalendTypeFilter.dynamicTypeCompatibleComponents) {
+            when(component.getName()).thenReturn(componentName);
+            assertTrue(Arrays.asList(new SparkBatchMetadataTalendTypeFilter(node).filter(types)).contains(DYNAMIC));
```
AlixMetivier, line 135: should not you rather test that the nodes that are compatible with dynamic is exactly SparkBatchMetadataTalendTypeFilter.dynamicTypeCompatibleComponents rather than making sure that all the compatible dynamic nodes contain dynamic ? It could be an empty list and still work ?

AlixMetivier (dismissed):
small remark for test case, lgtm

test/plugins/org.talend.core.runtime.test/src/org/talend/core/model/metadata/MetadataTalendTypeFilterTest.java
lbourgeois: Agree, seems functionally more logical even if test code coverage will be the same. Will try to modify that way

PR #4557, opened 2021-09-03
fix(TUP-32804): fix module index

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-32804


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/LocalLibraryManager.java
```diff
@@ -1468,6 +1468,16 @@ private void calculateModulesIndex(List<ModuleNeeded> modules, Map<String, Strin
             String moduleLocation = module.getModuleLocaion();
             // take maven uri from configuration to save in the index , don't generate by module name automatically
             String mavenUrl = module.getMavenURIFromConfiguration();
+            
+            if (StringUtils.isEmpty(mavenUrl)) {
+                mavenUrl = module.getMavenUri();
```
wchen-talend, line 1473: I think this call will cause issue.  the getMavenUri will call the index to guess the maven url if not configured , but at this point the index is not generated yet.
So I think we should only call the getMavenURIFromConfiguration to add maven urls configured from studio to the index.
For other libraries do not configured maven url , either get from the index, either generate a default one.

bhe-talendbj: revert this changed

zwxue (approved):

PR #4576, opened 2021-09-09
feat(TUP-32774): Notification when update is available.

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-32774


cmeng-talend (commented):
If the notication is only used by update, maybe can put it into featmanage.ui plugin? and use IStartup extension to trigger it

PR #4588, opened 2021-09-14
fix(TDI-46315):sap component and sap metadata have different module name/mvnurl.(#4567)

jzhao-talend:
https://jira.talendforge.org/browse/TDI-46315


vyu-talend (approved):
LGTM

PR #4602, opened 2021-09-17
fix(TDI-46727): Cannot pass a global variable of type date to a spark subjob using tRunJob after R2021-06

wwang-talend:


firefryy (approved):
lgtm

PR #4616, opened 2021-09-24
fix(TUP-33021):Feature Manager can't list any feature if studio install location has space.

hcyi:


main/plugins/org.talend.commons.runtime/src/org/talend/commons/utils/network/NetworkUtil.java
```diff
@@ -360,4 +362,21 @@ public static boolean isSelfAddress(String addr) {
         }
         return false;
     }
+
+    public static URI toUri(String path) throws Exception {
+        URI uri = null;
+        try {
+            String pattern = "^/?[A-z]:"; //$NON-NLS-1$
+            Pattern r = Pattern.compile(pattern);
```
nrousseau, line 370: regexp is slow... and seems was not used in other parts of the code... can it bring regressions?

and most strange thing... why set a toURI method in the NetworkUtil class..? which seems totally not related to this

cmeng-talend (approved):

PR #4659, opened 2021-10-09
fix(TDI-46572): add jboss dependency to .m2

pyzhou-talend:
https://github.com/Talend/tcommon-studio-ee/pull/2330


main/plugins/org.talend.librariesmanager/plugin.xml
```diff
@@ -62,6 +62,12 @@
                name="crypto-utils-@@CRYPTOUTILS_VERSION@@.jar">
          </library>
       </systemRoutine>
+      <systemRoutine
+            name="SerializeUtil">
+         <library
+               name="mvn:org.jboss.marshalling/jboss-marshalling/2.0.12.Final">
```
wwang-talend, line 68: seems here use maven path, should be "name"

main/plugins/org.talend.designer.maven.tos/resources/build/pom.xml
```diff
@@ -145,6 +145,11 @@
 			<artifactId>commons-compress</artifactId>
 			<version>1.21</version>
 		</dependency>
+        <dependency>
```
wwang-talend, line 148: better to align for format

main/plugins/org.talend.librariesmanager/plugin.xml
pyzhou-talend: @bhe-talendbj told me to do this. No idea why but Seems I have to use the mvn path. Using name to do the filter

main/plugins/org.talend.designer.maven.tos/resources/build/pom.xml
pyzhou-talend: done

main/plugins/org.talend.designer.maven.tos/resources/build/pom.xml
```diff
@@ -145,6 +145,11 @@
 			<artifactId>commons-compress</artifactId>
 			<version>1.21</version>
 		</dependency>
+		<dependency>
+			<groupId>org.jboss.marshalling</groupId>
+			<artifactId>jboss-marshalling</artifactId>
```
nrousseau, line 150: It's a LGPL jar.. we can't have it in the product by default, need to be in additional download if we really need it
Or best.. don't depend on LGPL jar at all..

nrousseau (approved):

PR #4792, opened 2021-11-19
(APPINT-34059) fix the following artifacts could not be resolved: org…

xldai:
….codehaus.plexus:plexus-utils:jar:3.3.0


dsergent (approved):
LGTM

PR #4813, opened 2021-11-25
APPINT-32136 Job & Run view displays DI job icon and/or label when a …

hweitalend:
…route or routelet editor is active (#4781)

- Add routelet icon


xldai (approved):
LGTM

PR #5111, opened 2022-03-10
fix(TDQ-19889): fix reload database functions for context prompt connection

msjian:


main/plugins/org.talend.metadata.managment.ui/src/main/java/messages.properties
```diff
@@ -546,3 +546,5 @@ HiveJDBCPropertiesDialog.title=Hive Properties
 HiveJDBCPropertiesDialog.desc=Hive properties configurations
 MetadataTalendTypeEditor.confirmTitle=Warning
 MetadataTalendTypeEditor.confirmMessage=Are you sure to replace current with default system settings?
+PromptDialog.title=Run Context {0} with parameters:
+PromptDialog.stringTip=The input text must be surrounded with single quote.
```
xqliu-talend, line 550: It is single quote, not double quote, right?

xqliu-talend (approved):

PR #5131, opened 2022-03-17
fix(TDI-47651): bump Redshift driver to 1.2.55.1083

ozhelezniak-talend:
https://jira.talendforge.org/browse/TDI-47651


undx (approved):
LGTM

PR #5204, opened 2022-04-11
Feat/tbd 13328

sponomarova:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/hadoop/version/ESynapseAuthType.java
```diff
@@ -0,0 +1,75 @@
...
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
+package org.talend.core.hadoop.version;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public enum ESynapseAuthType {
```
lbourgeois, line 18: Is ESynapseAuthType used in tcommon repo or can we move it to tbd repo as it is BD related ?

sponomarova: Moved it to bd side.
Pls, check in https://github.com/Talend/tbd-studio-se/pull/2587/files

lbourgeois (approved):
LGTM

PR #5205, opened 2022-04-12
update jackson bom

jiezhang-tlnd:


jiezhang-tlnd:
fixed on https://jira.talendforge.org/browse/TUP-35446

PR #5271, opened 2022-05-05
Temp commit

kjwang-talend:


main/plugins/org.talend.signon.util/src/main/java/messages.properties
```diff
@@ -0,0 +1,6 @@
+TMCRepositoryUtil.ap.displayName=Asia Pacific on AWS
+TMCRepositoryUtil.au.displayName=Australia on AWS
+TMCRepositoryUtil.eu.displayName=Europe on AWS
+TMCRepositoryUtil.us.displayName=United States - East on AWS
+TMCRepositoryUtil.us-west.displayName=United States - West on Azure
+SSOClientExec.error.timeout=Timeout waiting for login
```
wchen-talend, line 6: These data center are not expected to be translated. Can we have a json file to manage the data center instead of put them in the messages.properties and get displayname with hard coded stack key? It will be more easy to add or 
{"stacks":[{
 "ID": "cloud_us",
 "stack":"us",
 "display":"United States - East on AWS",
 "default": "true"
},
{.....
}
]}

PR #5343, opened 2022-05-31
fix(APPINT-34416) - update cxf version to 3.4.7

apoltavtsev:


xldai:
After TUP merged https://github.com/Talend/tcommon-studio-se/pull/5258 today,
This PR is unnecessary anymore, confirmed with Songbo.

PR #5348, opened 2022-06-02
chore(TDI-47843): update tck connectors to 1.27.10

bhe-talendbj:
https://jira.talendforge.org/browse/TDI-47843


pyzhou-talend (approved):
Thanks for the PR

PR #5426, opened 2022-07-04
fix(TDI-48203): [CVE]: routines.system.JSONObject(Object bean) use reflect to call method invoke, that's dangerous for evil script inject

wwang-talend:


coheigea (approved):

jzhao-talend (approved):
LGTM

PR #5440, opened 2022-07-06
fix(TUP-35745): Research: Bring a dark mode for Talend Studio as

zyuan-talend:
Experimental.


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/ui/view/RepositoryLabelProvider.java
```diff
@@ -480,6 +492,21 @@ public Color getBackground(Object element) {
...
+            IThemeEngine engine = manager
+                    .getEngineForDisplay(PlatformUI.getWorkbench().getActiveWorkbenchWindow() == null ? Display.getCurrent()
+                            : PlatformUI.getWorkbench().getActiveWorkbenchWindow().getShell().getDisplay());
+            ITheme curTheme = engine.getActiveTheme();
+            if (curTheme.getId().contains("dark")) {
+                return null;
+            }
+        } catch (Exception ex) {
+            ex.printStackTrace();
+        }
```
cmeng-talend, line 509: do we have another solution?

cmeng-talend (changes requested):

PR #5499, opened 2022-07-27
feat(TUP-35712):Track item signature with Pendo

jding-tlnd:
https://jira.talendforge.org/browse/TUP-35712


main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
```diff
@@ -2319,6 +2319,7 @@ public void logOnProject(Project project, IProgressMonitor monitor) throws Login
                 if (monitor != null && monitor.isCanceled()) {
                     throw new OperationCanceledException(""); //$NON-NLS-1$
                 }
+                PendoItemSignatureManager.getInstance().collectProperties();
```
cmeng-talend, line 2322: it will slow down the logon process, effect performance

cmeng-talend (changes requested):

main/plugins/org.talend.core.repository/src/main/java/org/talend/core/repository/model/ProxyRepositoryFactory.java
jding-tlnd: modified to make collect and send in another thread

PR #5557, opened 2022-09-01
feat(TUP-25445)acokit support context in repository metadata

jiezhang-tlnd:
https://jira.talendforge.org/browse/TUP-25445

### DIFF
--- a/main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/utils/TaCoKitConnectionContextUtils.java
+++ b/main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/utils/TaCoKitConnectionContextUtils.java
@@ -1,0 +1,116 @@
+// ============================================================================
+//
+// Copyright (C) 2006-2021 Talend Inc. - www.talend.com
+//
+// This source code is available under agreement available at
+// %InstallDIR%\features\org.talend.rcp.branding.%PRODUCTNAME%\%PRODUCTNAME%license.txt
+//
+// You should have received a copy of the agreement
+// along with this program; if not, write to Talend SA
+// 9 rue Pages 92150 Suresnes, France
+//
+// ============================================================================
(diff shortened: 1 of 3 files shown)


main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/utils/TaCoKitConnectionContextUtils.java
```diff
@@ -0,0 +1,91 @@
...
+import org.talend.designer.core.model.utils.emf.talendfile.ContextType;
+import org.talend.metadata.managment.ui.model.IConnParamName;
+
+
+public class TaCoKitConnectionContextUtils {
+
+	static List<IContextParameter> getContextVariables(final String prefixName, Connection conn,
+			Set<IConnParamName> paramSet) {
+		List<IContextParameter> varList = new ArrayList<IContextParameter>();
+		for (IRepositoryContextHandler handler : RepositoryContextManager.getHandlers()) {
```
cmeng-talend, line 35: the class is only used for tacokit, so seems no need to iterate for all other handlers? so seems the default [ExtendedNodeConnectionContextUtils.getContextVariables(label, conn, paramSet);](https://github.com/Talend/tcommon-studio-se/pull/5557/files#diff-05bd3fc39f401f98af544ae2892ef87da5b1de161b8e338a94b5b4115a2488dfR319) can handle it?

cmeng-talend (changes requested):
Please format the codes

main/plugins/org.talend.metadata.managment.ui/src/main/java/org/talend/metadata/managment/ui/utils/TaCoKitConnectionContextUtils.java
jiezhang-tlnd: yes. will update it

PR #5575, opened 2022-09-05
Revert "Route and data services features versions behave differently in Talend Runtime"

sbieliaievl:
Reverts Talend/tcommon-studio-se#5531


xldai (approved):
revert commit, LGTM

PR #5614, opened 2022-09-21
fix(APPINT-34581) - Studio cloud publish Talend Runtime versions

sbieliaievl:
fix(APPINT-34581) - Studio cloud publish Talend Runtime versions


apoltavtsev (approved):
LGTM

PR #5709, opened 2022-10-17
fix(TDI-48676):ResumeUtil OOM

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-48676


main/plugins/org.talend.librariesmanager/resources/java/routines/system/ResumeUtil.java
```diff
@@ -553,6 +553,16 @@ public void write(String content) throws IOException {
                 content = replace(content, "" + TextQualifier, "" + TextQualifier + TextQualifier);
             }
             
+            if (content.length() > 2 << 18) { //2M
```
wwang-talend, line 556: 2 << 18 appear 4 times, please remove the duplicate

main/plugins/org.talend.librariesmanager/resources/java/routines/system/ResumeUtil.java
```diff
@@ -553,6 +553,16 @@ public void write(String content) throws IOException {
                 content = replace(content, "" + TextQualifier, "" + TextQualifier + TextQualifier);
             }
             
+            if (content.length() > 2 << 18) { //2M
+                int index = 0;
+                for (; content.length() - index > 2 << 18; index += 2 << 18) {
+                    flush(true);
+                    final String substring = content.substring(index, index + (2 << 18));
```
wwang-talend, line 560: this will make lots of base 2M size memory middle size string, though not at the same time point, but make lots of press on jvm gc. Advice to use String.charAt method to read directly to avoid to make new array in background.

wwang-talend: public CharBuffer put(String src, int start, int end) api can help this

wwang-talend (changes requested):

main/plugins/org.talend.librariesmanager/resources/java/routines/system/ResumeUtil.java
pyzhou-talend: I cannot use Charbuffer here. As I'm using FileChannel to write the content to the log file. It requires the ByteBuffer (As the file needs encoding). 
First, this is an edge case, rere user will have large contexts like hundreds of MB.
Second, the substring should be recollected in Eden in GC, the GC in Eden is quite effective.

main/plugins/org.talend.librariesmanager/resources/java/routines/system/ResumeUtil.java
pyzhou-talend: I will change that

wwang-talend (approved):

PR #5834, opened 2022-11-24
fix(TDQ-20795):Upgrade Mysql 8 driver jar  [DON'T MERGE] 

qiongli:


msjian:
TUP-38935 already done

PR #5848, opened 2022-12-02
fix(TUP-37349)Data viewer for tCouchBaseInput get npe error

jiezhang-tlnd:
https://jira.talendforge.org/browse/TUP-37349


wwang-talend (approved):

main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/model/process/AbstractNode.java
```diff
@@ -509,7 +510,7 @@ public IElementParameter getElementParameter(String name) {
             }
             for (String key : childParameters.keySet()) {
                 IElementParameter param = childParameters.get(key);
-                if (param.getName().equals(name)) {
+                if (StringUtils.equals(name, param.getName())) {
```
cmeng-talend, line 513: The original null issue is not fixed

cmeng-talend (changes requested):

cmeng-talend (approved):

PR #5882, opened 2022-12-20
fix(TUP-37467):[8.0.1&7.3.1] custom component are no more build as

jding-tlnd:
snapshot but releases since june/july patch
https://jira.talendforge.org/browse/TUP-37467


main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/LocalLibraryManager.java
```diff
@@ -1324,6 +1324,46 @@ public Map<String, String> buildModulesIndexFromComponentAndExtension(IProgressM
...
+                    needDeploy = true;
+                } else {
+                    File jarFromMaven = new File(jarPathFromMaven);
+                    if (!jarFromMaven.exists()) {
+                        needDeploy = true;
+                    }
+                }
+
+                if (needDeploy) {
+                    File jarFile = new File(componentFolder, module.getModuleName());
```
nrousseau, line 1348: i would first try to find the jar name that matches mvnURI, then if not found, try with moduleName.
As we'll need to get rid of moduleName one day still. (this is just a duplicate info normally..)

cmeng-talend (approved):

nrousseau (changes requested):

main/plugins/org.talend.librariesmanager/src/main/java/org/talend/librariesmanager/model/service/LocalLibraryManager.java
jding-tlnd: ah ok I'll modify, thanks.

wchen-talend (approved):

PR #5888, opened 2022-12-22
feat(TDI-48882):revert codes that removed ingres/vectorwise

vyu-talend:
https://jira.talendforge.org/browse/TDI-48882


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/database/EDatabase4DriverClassName.java
```diff
@@ -33,6 +33,8 @@ public enum EDatabase4DriverClassName {
     IBMDB2ZOS(EDatabaseTypeName.IBMDB2ZOS, "COM.ibm.db2os390.sqlj.jdbc.DB2SQLJDriver"), //$NON-NLS-1$
 
     INFORMIX(EDatabaseTypeName.INFORMIX, "com.informix.jdbc.IfxDriver"), //$NON-NLS-1$
+    INGRES(EDatabaseTypeName.INGRES, "com.ingres.jdbc.IngresDriver"), //$NON-NLS-1$
```
jzhao-talend, line 36: why did you also revert the tup code?
you only need to add back the mapping files, no?

jzhao-talend (changes requested):

PR #5933, opened 2023-01-16
fix(TDI-49256) Upgrade tck in studio 7.3

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-49256


ozhelezniak-talend (approved):
lgtm

PR #6061, opened 2023-03-22
master/APPINT-35452 correct snapshot post fix

jillyan:
fix:
APPINT-35452  error org.osgi.framework.BundleException: Invalid Manifest header "Bundle-Version": 0.1.0-SNAPSHOT

the bundle version which will be used in Runtime should be 0.1.0.SNAPSHOT


arashankar (approved):
LGTM

PR #6118, opened 2023-04-13
TUP-38645 On dark mode, Job designer & subjob color can't restore to

kjwang-talend:
TUP-38645 On dark mode, Job designer & subjob color can't restore to correct default color
https://jira.talendforge.org/browse/TUP-38645


kjwang-talend:
Fixed in TUP-38656

PR #6244, opened 2023-06-14
fix(TBD-15341): Enable 'username' and 'password' only for LDAP auth in Impala compoents

sponomarova:
…n Impala


main/plugins/org.talend.repository.metadata/src/main/java/org/talend/repository/ui/wizards/metadata/connection/database/DatabaseForm.java
```diff
@@ -691,8 +691,11 @@ public void initialize() {
             urlConnectionStringText.setText(getConnection().getURL());
 
         }
-        usernameText.setText(getConnection().getUsername());
-        passwordText.setText(getConnection().getRawPassword());
+        if (!isDBTypeSelected(EDatabaseConnTemplate.IMPALA)) {
+            usernameText.setText(getConnection().getUsername());
```
lbourgeois, line 695: Here it is about setting the text but there is no call to show() or hide() methods so I am a bit confused.

PR #6279, opened 2023-06-29
fix(TDI-49933): snowflake metadata time issue

pyzhou-talend:
https://jira.talendforge.org/browse/TDI-49933


jzhao-talend (approved):
LGTM

PR #6285, opened 2023-07-05
fix(TUP-39291): Remove Talend Exchange link from Studio.

zyuan-talend:


main/plugins/org.talend.core.runtime/src/main/java/org/talend/core/PluginChecker.java
```diff
@@ -268,7 +268,7 @@ public static boolean isMDMPluginLoaded() {
     }
 
     public static boolean isExchangeSystemLoaded() {
-        return isPluginLoaded(EXCHANGE_SYSTEM_PLUGIN_ID);
+        return false;
     }
```
wchen-talend, line 273: better if do cleanup and deleted this public function directly ?

zyuan-talend: ok, will implement it in https://jira.talendforge.org/browse/TUP-39605.

jding-tlnd (approved):

PR #6443, opened 2023-10-08
feat(TDQ-21334):Support oracle custom ssl for DQ

qiongli:


main/plugins/org.talend.metadata.managment/src/main/java/org/talend/metadata/managment/utils/MetadataConnectionUtils.java
```diff
@@ -536,7 +537,7 @@ public static boolean isOracleCustomSSL(Connection connection) {
             if (EDatabaseTypeName.ORACLE_CUSTOM.getDisplayName().equals(dbConn.getDatabaseType())) {
                 if (EDatabaseVersion4Drivers.ORACLE_12.name().equals(dbConn.getDbVersionString()) || EDatabaseVersion4Drivers.ORACLE_18.name().equals(dbConn.getDbVersionString())) {
                     
-                    return true;
+                    return Boolean.parseBoolean(dbConn.getParameters().get(ConnParameterKeys.CONN_PARA_KEY_USE_SSL));
```
bhe-talendbj, line 540: It seems that by default it is always true before your change. With your change, default it is false now. So suggest 

if(dbConn.getParameters().get(ConnParameterKeys.CONN_PARA_KEY_USE_SSL)==null){
return true;
} else {
return Boolean.parseBoolean(dbConn.getParameters().get(ConnParameterKeys.CONN_PARA_KEY_USE_SSL));
}

qiongli: The method name isOracleCustomSSL,  if  'ConnParameterKeys.CONN_PARA_KEY_USE_SSL'  is null, I think it should return false

qiongli:
close it  and create a new one #6447

PR #6487, opened 2023-10-20
APPINT-35995: [801]Route with tHttpclient failed to run with docker image when build by CI

apoltavtsev:


arashankar (approved):
LOOKS GOOD TO ME!

PR #6607, opened 2023-12-07
chore(TUP-40992): fix exclude group for copy-dependencies

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-40992


nrousseau:
merged to TUP-41109 manually

PR #6639, opened 2023-12-16
fix(TUP-40887): Mac Sonoma issue with image flipped

hzhao-talendbj:


main/plugins/org.talend.rcp/src/main/java/org/talend/rcp/TalendSplashHandler.java
```diff
@@ -30,4 +34,21 @@ public TalendSplashHandler() {
             instance = this;
         }
     }
+
+    @Override
+    public IProgressMonitor getBundleProgressMonitor() {
+        IProgressMonitor bundleProgressMonitor = super.getBundleProgressMonitor();
+        if (bundleProgressMonitor instanceof ProgressMonitorPart) {
+            ProgressMonitorPart pmp = (ProgressMonitorPart) bundleProgressMonitor;
+            Image backgroundImage = getSplash().getShell().getBackgroundImage();
```
cmeng-talend, line 43: it's better to add more null pointer check here, in some cases, the splash image may be missing when starting studio, in these cases, not sure whether they'll still have value or null? maybe can simply remove the splash path from config.ini, and test the behavior here

main/plugins/org.talend.utils/src/main/java/org/talend/utils/image/ImageUtils.java
```diff
@@ -0,0 +1,69 @@
...
+        transform.dispose();
+
+        return newImage;
+    }
+
+    public static boolean isSonomaOrLater() {
+        boolean result = false;
+        try {
+            result = Platform.OS_MACOSX.equals(Platform.getOS())
+                    && ("14.0".compareTo(System.getProperty("os.version", "0")) <= 0);
```
cmeng-talend, line 47: does mac only contains two digits for the version? maybe need to compare the first digit only?

cmeng-talend (approved):

nrousseau (approved):

PR #6657, opened 2023-12-21
chore(TUP-40272): Try to find way to avoid m2e.core.ui issue

bhe-talendbj:
https://jira.talendforge.org/browse/TUP-40272


kjwang-talend (approved):
LGTM

